Issue #47, opened 2014-10-30
Last field in a record is not parsed when it does not have brackets nor comma


gpoo:
In BibTex is valid that fields like _month_ or _year_ are not braced.  However, when these are the last field in a record, they are not parsed.

Test case with both, month and year:

``` python
from bibtexparser.bparser import BibTexParser

if __name__ == '__main__':
    bibtex = """@ARTICLE{Cesar2013,
      author = {Jean César},
      title = {An amazing title},
      month = jan,
      year = 2013,
    }

    @ARTICLE{Cesar2014,
      author = {Jean César},
      title = {An amazing title},
      month = jan,
      year = 2014
    }

    @ARTICLE{Cesar2012,
      author = {Jean César},
      title = {An amazing title},
      year = 2012,
      month = jan
    }

    @ARTICLE{Cesar2011,
      author = {Jean César},
      year = 2011,
      title = {An amazing title},
      month = jan,
    """

    with open('/tmp/bibtex.bib', 'w') as bibfile:
        bibfile.write(bibtex)

    with open('/tmp/bibtex.bib', 'r') as bibfile:
        bd = BibTexParser(bibfile.read())

    for entry in bd.get_entry_list():
        print([entry[x] for x in sorted(entry)])
```

The output is:

``` python
['Jean César', 'Cesar2013', 'jan', 'An amazing title', 'article', '2013']
['Jean César', 'Cesar2014', 'jan', 'An amazing title', 'article']
['Jean César', 'Cesar2012', 'An amazing title', 'article', '2012']
['Jean César', 'Cesar2011', 'jan', 'An amazing title', 'article', '2011']
```

The expected  output is:

``` python
['Jean César', 'Cesar2013', 'jan', 'An amazing title', 'article', '2013']
['Jean César', 'Cesar2014', 'jan', 'An amazing title', 'article', '2014']
['Jean César', 'Cesar2012', 'jan', 'An amazing title', 'article', '2012']
['Jean César', 'Cesar2011', 'jan', 'An amazing title', 'article', '2011']
```

sciunto:
This can be check once #98 is solved.

vext01:
I've just hit this problem too. Any plan to merge any time soon?

sciunto:
@vext01 Any help to make progress on those issues is appreciated :)

vext01:
This is not fixed in master, as the last month field will not parse. See #152.

Phyks:
Replacing
```
[...]

PR #54, opened 2014-12-17
[WIP] Regression comments starting with % and no final coma

sciunto:
See #45 for a full discussion.

This branch integrates a test that reproduces the issue. The test is skipped and must be addressed.

Issue #43, opened 2014-09-25
"comment" lines without leading @ breaks parser


auxym:
I'm not sure if such lines conform to bibtex syntax, but they are found in standard packages. See, for example, [IEEEabrv.bib](http://www.tex.ac.uk/CTAN/biblio/bibtex/contrib/IEEEtran/IEEEabrv.bib).

Some minimal examples:

```
@STRING{Foo = "bar"}
This is a comment
This is a second comment.
```

This returns the incorrect dictionary entry `OrderedDict([('foo', 'bar"} This is a comment This is a second comment.')])`

```
@STRING{Foo = "bar"}
This is a comment
STRING{Baz = "This should be interpreted as comment."}
```

This raises an exception when parsed.

I had a quick look at the parser code. Maybe, instead of using a new `@` line as the record delimiter, it should use the braces?

As a workaround, I'm removing all lines not starting with `@` before parsing the aforementioned IEEEabrv.bib fil, but this only works because this particular file only has one-liners.

sciunto:
Fixed in master. Thanks.

PR #64, opened 2015-05-07
New parser based on pyparsing.

omangin:
Hi,

Looking for a fix for issue #47 (that at some point disappeared and then came back), I found that the current parser was quite difficult to modify and its behavior difficult to understand. I thus implemented an other parser based on [pyparsing](https://pyparsing.wikispaces.com/), that enables the definition of the parser through its grammar.

This comes with a little overhead but the resulting parser is greatly simplified and in my opinion easier to maintain because its behavior (and the corresponding implementation choices) are clearer.

I found it quite difficult to find good references on the bibtex format (which seems to be mainly reverse-engineered from the bibtex binary), however since [maverick.inria.fr/~Xavier.Decoret/...](http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html) was cited in the tests and documentation, I based my work on it.

So, the new parser is completely integrated in the Bibtexparser class and the switch should be transparent. It also pass all tests but three, on which some choices have to be taken that I wanted to first submit to the community. In order to make it pass the other tests, I had to fix some behaviors as detailed below.

- I updated the tests to remove unquoted values such as `jan`, except if they are meant to be interpreted as strings.

- In particular common strings are often pre-defined in bibtex such as month short names. I thus included a mechanism to add such common strings in `bibdatabase.py` although it is not active by default.

- Similarly `bwriter.py` has been updated so that preambles are quoted (otherwise it conflicts with the use of strings in preambles).

- The handling of strings and their interpolation has changed since string definitions and expressions are now parsed correctly.

- Maybe the definition of standard types (that I moved to a global variable for clarity) should go to `bibdatabase.py`.

- In the current implementation, `@comment` are handled by selecting everything between the braces. It seems that it differs from [what Decoret describes](http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html#comment) (as being bibtex' behavior), that is to only include what is between `@comment` and the end of current line. I implemented the latter behavior but left two tests in `test_comments.py` as they were (this makes two of the three non-passing tests). Which behavior should be kept?

- In `test_traps` the newline between author names is removed. This is a strange expected behavior at the bibtex parser level that do not occur any more (but I left the non-passing test). Authors are correctly handled by the author customization function later.

A last comment on strings. I kept the previous behavior of interpolating strings on the go. Anyway it may be beneficial to represent bibtex strings with a custom object so that interpolation is delayed which would enable bwriter to declare again the parsed strings.


sciunto:
This just looks awesome! Using pyparsing is definitely the right way to handle the problem and it was on my list.

Travis is complaining because pyparsing must be install first. Are you familiar with travis? Basically, you need to fix .travis.yml and the test suite will be executed.

Thanks!

sciunto:
PS: I'll look in more details asap.

sciunto:
Two quick comments:
* We need docstrings for each functions (included private or internal) with a description of the variables. Sometimes, it looks obvious but it helps to maintain the code.
* I would like to keep the logger (what's the code is doing, what's the content of the main variables). For unexpected behaviors reported by users, it's helpful to understand the error.

omangin:
Thanks for your comments!

- About the docstrings, I will take some time to document the new functions (which are not so many).
- For the logger part I will have to see how it can integrate with pyparsing.
- Updating travis.yml should be quick too.

Then we will need to decide what to do with the few *gray zones* mentioned above.

I will let you know when the first steps are done.

sciunto:
I also had difficulties to find a proper description of bibtex files.
Some choices in the test suite might be poor, and it could be changed if there is obviously an enhancement.

I'm in favor to move the standard types in bibdatabase.py


About the @comment. The current behavior comes from a contributor. For my usage, I do not care about comments.
Bibtex also don't, so the problem does not exists for bibtex.

But, some users want their comments to be parsed. And there, a choice beyond the standard must be made.
One question for which I do not have a clear answer is: How do users want to have their comments parsed?
I believe that most users ignore that @comment does not require an ending }. Thus, I think that the current behavior is probably closer to the majority of usecases even if it's not standard. However, I'm ready to hear counter-arguments.
BTW, these issues #43 and #54 could be tackled at the same time.

About authors, I'm not happy with the current behavior. Another python project (Pybtex I think) that I discovered recently handles that in a better way.


You can add your name to the contributor.txt file. :)


thanks!

sciunto:
hi @omangin Did you have a chance to look at this pr ?

Thanks ;)

omangin:
So, I had some time work back on the new parser. I re-organized the code by putting all the pyparsing expressions in a separate file. 

Now the tests pass on both python 2 and 3. Some things are still incomplete regarding some special cases of quotes / braces combinations with escapes. But I am not sure these cases were handled by the previous code anyway.

One issue, though, is that the pyparsing code seems to be slower than the original code. I am not sure whether this is a big issue. This may depend on users.
 
It would be great now to get some testing from other users on the code.

gpoo:
@omangin How do yo measure the slowness?

omangin:
@gpoo First the test are now slower (although I added some but this does not account for the difference). The other way is simply to time the parsing time for a big bibtex file.

gpoo:
@omangin I wonder about this remark:

> * I updated the tests to remove unquoted values such as jan, except if they are meant to be interpreted as strings.

The same specification you link states that an entry like:  `month = jan,` is valid. From what I understand from your comment, this won't work. (Well, it does not parse bibtex files with that format either :-)  What is the rationale to not parse it?

gpoo:
I forgot to add, that Python-Bibtex (http://sourceforge.net/p/pybliographer/python-bibtex-1.2/ci/master/tree/) has a parameter `strict' that might be something to think about.  If you set `strict` then complain that `jan` is not defined. Otherwise, be more lenient with those files for this particular case(s).

gpoo:
The default (plain) Bibtex defines the abbreviated month as macros. Therefore, it is valid to write `month = jan`.

Even more, the documentation says:

> It’s best to use the three-letter abbreviations for the month, rather than spelling out the month yourself. This lets the bibliography style be consistent. And if you want to include information for the day of the month, the month field is usually the best place. For example
> 
>  `month = jul # "~4,"`
> 
> will probably produce just what you want.

You can check it in the PDF available from the BibTex package http://www.ctan.org/tex-archive/biblio/bibtex/base/

About the formatting, it might worth to make it compliant with what says http://mirrors.ctan.org/biblio/bibtex/base/btxbst.doc (it is a text plain file).

omangin:
OK. Thanks @gpoo  for your feedback. I had not looked that far in the question. So I believe the conclusion is that we have to accept common strings by default. (That is `common_strings=True` as the default value for the parser.)

Then comes a new question. Currently the new parser uses a special object from `bibdatabase.py` called `BibDataString` to represent parsed string object. This, in theory, enables to delay the interpolation of bibtex strings and chose, for example, to later write it as such in a new file. It seems that the current parser by default interpolates strings and that the default interpolation content for undefined strings is the key. This means that `month = jan` is parsed into `'month': 'jan'` which would be exported as `month = {jan}` by the writer. Then I guess bibtex itself would not interpolate "jan" since it is not parsed as a string name any more.

What is yet implemented for the new parser is to interpolate known strings ("known" meaning previously defined in the bibtex file or being a common string if the option is set). This is close to the previous behavior but not exactly the same since unknown strings raise errors.

So here are the options to implement:
- the clever way: represent string expressions as such (lists of python `string` and `BibDataString`). This unables the interpolation as well as the export in the original form. This however breaks the compatibility with the current behavior. (Although it is possible to hack these *bibtex string expressions* to behave as python string.) 
- the always working solution: always interpolate. This would be fully compatible with previous behavior with the aforementioned issue.
- the strict option: fail on undefined strings, either interpolate or keep the full structure.

I think that the interpolation fits well in as a customization function were I should move it. The other solutions are all easy to implement but might break some compatibility. What do you think should be the default?

sciunto:
Just to say I'm nearby :) It looks great. i'll try to look at it during the holidays.

sciunto:
@omangin Could you rebase please? I just  release a new version based on the historical parser. I integrated some skipped unittests for which we can try your implementation.

Thank you.

sciunto:
@omangin  I made a PR : https://github.com/omangin/python-bibtexparser/pull/1

sciunto:
@omangin I have few comments for now:

* We need to pass the test I defined in test_bparser.py  called test_article_no_braces. The previous parser handled that case even if no test was made on it. You need to remove the decorator to enable it.
* Could you clarify a little the different steps (with comments) involving pyparser. For now, it's difficult to identify what's doing what if we don't know how pyparsing works. On my side, I need to read their documentation. It's mainly about the class BibtexExpression, __init__() function.

I need to look at your last comment and then, I think we are nearly ready to merge. :)

omangin:
@sciunto 
- Thanks for the test. I un-skipped it and fixed the corresponding bug.
- I also adds comments to the expressions. I hope it will be clearer now.

sciunto:
Thank you so much again! I opened #98 about the open question. I merge this PR, so it's not blocking for other contributions.

PR #73, opened 2015-08-02
Fix unicode write

Erotemic:
Writing a bibtex file previously failed if there were unicode characters in any entry fields.


sciunto:
Thanks!

PR #82, opened 2015-10-09
Extract protect_uppercase from homogeneize_latex_encoding

geritwagner:
Makes it possible to use homogeneize_latex_encoding without necessarily executing protect_uppercase (useful if protect_uppercase is not needed, buggy, or if the bib-file is too big to check the changes).


sciunto:
Fixed in #126

Issue #69, opened 2015-07-17
One-line bibtex records are not supported


wetneb:
Some sources (such as ORCiD) return bibtex records with no newlines between the fields:
<pre>
@article { cortadella1993,title = {Division with speculation of quotient digits},journal = {Proceedings - Symposium on Computer Arithmetic},year = {1993},pages = {87-94},author = {Cortadella, Jordi and Lang, Tomas}}
</pre>
It seems that the parser does not support this format.

stuz5000:
I've also run into this. Is making python-bibtexparser unusable for me. Was there a workaround (such as adding line breaks)?

sciunto:
It's not supported and the way the parser is written doesn't make this task easy to tackle nicely.
I hope to be able to work on #64 in few months.

sciunto:
TODO: check if #64 solves this bug.

sciunto:
Fixed by #64.

PR #89, opened 2015-10-22
[WIP] Handle one line records

sciunto:
See #69


sciunto:
Merge only the test. Let's check if #64 solves the issue

Issue #117, opened 2016-02-19
can not parse Journal field


rudrab:
Hi,
I am trying to parse a bibtex file, and it is parsing most field, but not journal:
```
>>> import bibtexparser
>>> with open("report.bib") as bibfile:
...     bibtex_str = bibfile.read()
... 
>>> bib_database = bibtexparser.loads(bibtex_str)
>>> print(bib_database.entries)
[{'pages': '1', 'title': 'First-principles calculation of the Gilbert damping parameter via the linear response formalism with application to magnetic transition metals and alloys', 'ENTRYTYPE': 'article', 'author': 'Mankovsky, S. and K{\\"o}dderitzsch, D. and Woltersdorf, G and Ebert, H.', 'volume': '87', 'year': '2013', 'ID': 'ebert2013'}, {'pages': '066603', 'title': '\\textit{Ab Initio} Calculation of the Gilbert Damping Parameter via the Linear Response Formalism', 'ENTRYTYPE': 'article', 'author': 'Ebert, H. and Mankovsky, S. and K{\\"o}dderitzsch, D. and Kelly, P. J.', 'volume': '107', 'year': '2011', 'ID': 'ebert2011', 'publisher': 'American Physical Society', 'month': 'Aug'}]
```

The report.bib is:
```
@article{ebert2013,
	Author="Mankovsky, S. and K{\"o}dderitzsch, D. and Woltersdorf, G and Ebert, H.",
	Year="2013",
        Title="First-principles calculation of the Gilbert damping parameter via the linear response formalism with application to magnetic transition metals and alloys",
	Volume="87",
	Pages="1",
	Journal="Phys Rev B"
}
@article{ebert2011,
	Author="Ebert, H. and Mankovsky, S. and K{\"o}dderitzsch, D. and Kelly, P. J.",
	Year="2011",
	Title="\textit{Ab Initio} Calculation of the Gilbert Damping Parameter via the Linear Response Formalism",
	Publisher="American Physical Society",
	Volume="107",
	Pages="066603",
	Month="Aug",
	Journal="Phys. Rev. Lett."
}

```
You can see Journal value Phys Rev B or Phys. Rev. Lett is not printed.
I am using 0.6.2 of bibtexparser with python 3
```
pip3 list|grep bibtex
bibtexparser (0.6.2)
```

sciunto:
Thank you @rudrab for your bug report. This already fixed in master. Your bibtex revealed another bug in master and I made a bug report.

[...]

PR #118, opened 2016-02-19
Test failure if title contains {\"u} et no outer { }

sciunto:
Cc @omangin 

From #117.


omangin:
OK. This should fix it.

https://github.com/omangin/python-bibtexparser/commit/c5bd18982f87f94b7836cc88b5964b87878f6ed2

sciunto:
See #120

PR #120, opened 2016-03-14
Fixes test failure if title contains {\"u} et no outer { }

omangin:
See  #117 and #118.


sciunto:
Merged! Thanks @omangin

Issue #138, opened 2016-05-24
few enhancements to decide based on #137 (name features)


sciunto:
From #137 
- [ ] Mention that in the tutorial?
- [ ] decide what we do about `customization.getnames()` and related tests.
- [ ] review #9 
- [ ] update CHANGELOG

bcbnz:
#9 should be fixable once pull request #140 is sorted (I did some renaming etc there). I'll take a look at it in the next few days.

The customization.author() and customization.editor() functions can also be rewritten in terms of the utils.splitnames() added in #140 to make them more robust. I'll also have a go at that.

I've updated the CHANGELOG to mention my additions in #140.

PR #140, opened 2016-06-10
Add utils module with splitnames() and parsename() functions

bcbnz:
Firstly, this pull request adds a `utils.splitnames()` function to break apart a string `"Donald E. Knuth and Leslie Lamport"` into a list of individual names `["Donald E. Knuth", "Leslie Lamport"]`. This obeys BibTeX brace level rules so `"{Simon and Schuster}"` results in an output of `["{Simon and Schuster}"]`. It also handles a leading or trailing 'and' in the same way as BibTeX, treating it as part of the name so that `"John Smith and Phil Holden and"` becomes `["John Smith", "Phil Holden and"]`. There are a number of test cases for this function.

Secondly, the `customization.splitname()` function I previously submitted is renamed to `utils.parsename()`. I feel this is better suited to being in the utils module. The name change better reflects what the function does, and also avoids the confusion of having `splitname` and `splitnames` functions.

Finally, two scripts intended for developer use are added to a new top-level `misc` folder. The `misc/splitnames.py` script shows how BibTeX splits a given string of names, and the `misc/parsename.py` script shows how it parses a given name. They take input from the user (either as commandline arguments, or interactively if run with no arguments), then run BibTeX with a custom style file to generate the output. I used these to generate test cases for the two corresponding functions, and they may be useful for the same purpose in the future. I arbitrarily named the folder `misc`, if you want them in a different location let me know.


sciunto:
Thanks!  I try to read it and make comments asap :)

sciunto:
I think it would be better to rename misc in something like "testcase_generator". In that directory, I would add a README.rst file to explain the existence of this folder (just what you said above).

just for curiosity, have you considered to use pyparsing for some parts of your functions?
It makes things more compact and generally less buggy. Nevertheless, I think this function is far enough of the lib core, so it's OK for me.

Would you like to cover this with few lines in the documentation? Also, how would you like to articulate this with `getnames()` and `author()`? I would prefer to treat them here also. I'm not totally sure about creating utils.py. I think that all actions on bibtex field contents must be viewed as a customization. I view on this is basically to keep two blobs: a core to parse the general structure of a bibtex file, and a shell to modify the content and its structure (string vs list). This way, I believe it's easy to explain to users.

Apart these little comments, that's a great and valuable contribution. Thanks!

Phyks:
@bcbnz any news on this (and on https://github.com/sciunto-org/python-bibtexparser/issues/138)? Would be awesome to get this merged, it is one of the remaining [issue to sort out before next release](https://github.com/sciunto-org/python-bibtexparser/milestone/1).

Thanks!

sciunto:
@Phyks Feel free to update the branch if you would like. It's quite a long time that this PR has not been updated.

Phyks:
Ok, I'll try to have a look as soon as possible :)

PR #151, opened 2016-11-17
Add author parsing code.

ChrisBarker-NOAA:
So I added a bunch of author parsing code, and only then discovered that 

customization.splitname() already existed.

And frankly, it's less kludgy than my code.

However, I ran all my tests against it, and there is a failure of an example from The LaTeX Companion.

I don't really think this code should be merged -- splitname() is a lot less kludgy -- but I thought I'd post it if folks are interested. 

I do think that we should fix splitname() to do the right thing with my tests -- and I also don't like the splitname API much

So I may submit a PR for that at some point.

### DIFF
--- a/bibtexparser/author_parse.py
+++ b/bibtexparser/author_parse.py
@@ -1,0 +1,247 @@
+#!/usr/bin/env python3
+
+
+"""
+This script merges a new bib file with an existing one.
+
+It is not very smart -- mostly what it does is give a "Chris Style" key.
+
+Then , with the merge it will make sure you don't have duplicate keys --
+but it will allow duplicate entries!
+
+We can make that smarter, later, but duplicates can be hard to identify!
+
+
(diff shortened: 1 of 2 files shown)


lfdversluis:
I like this functionality. No promises, but I may have a look at it unless something was added to address the issue already (@sciunto @omangin)?

omangin:
Thanks for the contribution. I have not looked the details so much but:
– please contribute the tests for the case that `splitname` does not currently covers (even if you don't feel like fixing the function right now it is fine to contribute a test that would be skipped for now with some attached issue),
– I guess the API change is about using the `namedtuple`; that something that can be discussed here, if this is something that users deem useful it can be implemented as an alternative (it is pretty straightforward to go from one to the other or we could make an object that supports both by forwarding `__getitem__` to the `__getattribute__`).

PR #154, opened 2017-03-22
Fix `convert_to_unicode` for unbraced sequences

### LINKED ISSUE #121: convert_to_unicode
I am using the 

parser.customization = convert_to_unicode

feature, but in the utf-8 output the accents are coming out on the character before the correct one.  What would cause this?

Phyks:
`convert_to_unicode` customization was converting `\"u` (without braces)
to `\u0308u` which was supposed to be an `u` letter with an umlaut.
However, the `\u0308` unicode character is supposed to be *after* the
letter, contrary to the LaTeX encoding of the accent.

This commit fixes it and additionally wrap the `convert_to_unicode`
customization in a dedicated function that could be reused.

Closes #121.

### DIFF
--- a/bibtexparser/customization.py
+++ b/bibtexparser/customization.py
@@ -7,11 +7,10 @@ You can find inspiration from these functions to design yours.
 Each of them takes a record and return the modified record.
 """
 
-import itertools
 import re
 import logging
 
-from bibtexparser.latexenc import unicode_to_latex, unicode_to_crappy_latex1, unicode_to_crappy_latex2, string_to_latex, protect_uppercase
+from bibtexparser.latexenc import latex_to_unicode, string_to_latex, protect_uppercase
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sciunto:
@Phyks Thanks! Please, check travis

```
======================================================================

FAIL: test_accent (test_latexenc.TestLatexConverter)

----------------------------------------------------------------------

  File "/home/travis/build/sciunto-org/python-bibtexparser/bibtexparser/tests/test_latexenc.py", line 29, in test_accent

    self.assertEqual(result, expected)

AssertionError: u'{\\`a} {\\\'e} {\\`e} \\"o' != u'{\\`a} {\\\'e} {\\`e} {\\"o}'

- {\`a} {\'e} {\`e} \"o

+ {\`a} {\'e} {\`e} {\"o}

```

Phyks:
Indeed, I was waiting for Travis tests to finish to be sure about the PR. I just noticed there is also the same issue as #121 with other accents such as `\~`.

I think the easiest way to proceed would actually be to simply convert TO NFD unicode form https://docs.python.org/2/library/unicodedata.html#unicodedata.normalize (with composition of character and accents) and eventually use `unicodedata` afterwards to normalize everything. This way, we should not have to worry about position of spaces, braces and so on. I'll update my patch accordingly ASAP.

Phyks:
This updated commit should do the trick. It passes all unittests on Python 2 and 3 on my machine. LaTeX to Unicode conversion is still fragile, but enhancing it much more would basically requires writing a full LaTeX parser, which is not the main point of this library I think (plus other tools such as Pandoc handles it very well).

Phyks:
Hmm, looks like I had some troubles as I worked on this with two different computers, and I messed up with the commit, sorry :/ I'll push it again tonight.

PR #155, opened 2017-03-27
Fix `convert_to_unicode` for unbraced sequences

### LINKED ISSUE #121: convert_to_unicode

Phyks:
`convert_to_unicode` customization was converting `\"u` (without braces)
to `\u0308u` which was supposed to be an `u` letter with an umlaut.
However, the `\u0308` unicode character is supposed to be *after* the
letter, contrary to the LaTeX encoding of the accent.

This commit fixes it and additionally wrap the `convert_to_unicode`
customization in a dedicated function that could be reused.

Also makes the fix for every other combining diacritic mark.

Closes #121.

@sciunto Sorry for the mess with the previous PR, I just got mixed up with the state of my fork on my two laptops :/ This PR should pass all the tests (at least it does on my laptop). I am waiting for Travis feedback.

### DIFF
--- a/bibtexparser/customization.py
+++ b/bibtexparser/customization.py
@@ -7,11 +7,10 @@ You can find inspiration from these functions to design yours.
 Each of them takes a record and return the modified record.
 """
 
-import itertools
 import re
 import logging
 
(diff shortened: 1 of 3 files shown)


sciunto:
Looks good to me. Thanks!

Phyks:
Thanks!

PR #158, opened 2017-03-29
Customization to strip fields of `{`, `}` etc

### LINKED ISSUE #116: Strip title of "{", "}" and similar
I've just tried bibtexparser for https://github.com/MartinThoma/algorithms/tree/master/bib2html and I noticed that `entry.title` can be something like `On-line {Recognition} of {Handwritten} {Mathematical} {Symbols}.`.

Is it possible to strip those / to make the title similar to what it gets rendered to?

Phyks:
Add a `add_plaintext_fields` customization to strip fields of braces and
similar characters, getting only the plaintext. It processes every
available fields and for each field, it adds a `plain_*` field
containing the plaintext version.

Should close #116.

### DIFF
--- a/bibtexparser/customization.py
+++ b/bibtexparser/customization.py
@@ -10,13 +10,15 @@ Each of them takes a record and return the modified record.
 import re
 import logging
 
+from builtins import str
+
 from bibtexparser.latexenc import latex_to_unicode, string_to_latex, protect_uppercase
 
 logger = logging.getLogger(__name__)
 
 __all__ = ['splitname', 'getnames', 'author', 'editor', 'journal', 'keyword',
            'link', 'page_double_hyphen', 'doi', 'type', 'convert_to_unicode',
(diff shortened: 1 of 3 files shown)


sciunto:
Awesome! thanks.

Phyks:
Yw

pbelmans:
One issue with this is that it will also replace curly brackets inside math expressions. Some papers in mathematics contain both math expressions with curly brackets, and curly brackets for capitalisation reasons. The current situation breaks the mathematics.

I'm not saying this is is a common issue, I just wanted to point it out.

Phyks:
Indeed… good catch! I think we can do something better with a regex or listing the things we actually want to transform (see homogenize latex encoding). This, or the doc should emphasize the issue.

Could you open an issue to track this? Also, if by chance you can have a fix, feel free to open a PR!

Thanks!

PR #183, opened 2017-11-02
fixed problem that BOM recognition doesn't actually work

Palpatineli:
BOM detection didn't work in python3 so bib files with utf-8-bom was not read into database.

### DIFF
--- a/bibtexparser/bparser.py
+++ b/bibtexparser/bparser.py
@@ -195,12 +195,14 @@ class BibTexParser(object):
 
     def _bibtex_file_obj(self, bibtex_str):
         # Some files have Byte-order marks inserted at the start
-        byte = '\xef\xbb\xbf'
-        if not isinstance(byte, ustr):
+        byte = b'\xef\xbb\xbf'
+        if isinstance(bibtex_str, ustr):
             byte = ustr(byte, self.encoding, 'ignore')
-        if bibtex_str[:3] == byte:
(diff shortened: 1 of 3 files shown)


Phyks:
Hi,

I had a quick look, but cannot do a full review right now. I'll have a deeper look at it ASAP. Few questions meanwhile:

* Did you check your code is working fine with both Python 2 and Python 3? I see you address a Python3 issue, but we shouldn't break Python2 compatibility.
* As far as I understand, this is a bugfix for BOM detection. Could you add a matching test in our testsuite to validate the fix and ensure this issue never comes back?

Thanks

Palpatineli:
I have added two tests for this: appears that the issue only appears in with python3. in python2 open() defaults to open() give a byte string anyway and was correctly handled. Is there a way I can update this pull-request, or should I close this one and open another one?

Phyks:
> Is there a way I can update this pull-request, or should I close this one and open another one?

The PR has have auto-updated when you pushed to your `master` on your fork.


Thanks for the tests and the extra infos! I'll try to have a deeper look ASAP, probably by tomorrow evening.

Phyks:
I had a chance to give a better look and it looks good for me, thanks! @Palpatineli could you squash all your commits in a single one please?

Palpatineli:
I have squashed the commits into one. Thanks for your time looking at it.

Phyks:
Thanks for your contribution!

Issue #143, opened 2016-09-20
biblatex entry type


rudrab:
Hi,
while its working well for bibtex, it can't parse biblatex's entry (e.g. @online).
Do you have any plan to add biblatex's features? 
This is possibly same as #105 . So, in this case, do you have any idea when 0.8 is going to release?

lfdversluis:
I support this feature. I was surprised to find out just now `@online` are silently (!) ignored.

omangin:
Hi @rudrab. Thank you for that.

Non-standard bibtex entries are indeed ignored by default. Which is actually not what the documentation of the `BibtexParser` class says. (That's a first thing to fix.) You can change the behavior through the `ignore_nonstandard_types` argument.

I am surprised that it does so silently for you. I just tested it and get a python warning. What version of python and bibtexparser are you using?

Finally, we should probably have two sets of standard entry types for bibtex and biblatex.

PR #189, opened 2018-02-10
Minor fixes

omangin:
– better test coverage for the `ignore_nonstandard` parameter (see #143),
– more documentation on `BibtexParser` parameters,
– small cleanup in `test_bparser`.

### DIFF
--- a/bibtexparser/tests/test_bparser.py
+++ b/bibtexparser/tests/test_bparser.py
@@ -54,13 +54,6 @@ def customizations_latex(record):
 
 class TestBibtexParserList(unittest.TestCase):
 
-    def test_wrong(self):
-        """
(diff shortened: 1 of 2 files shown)


bibtexparser/tests/test_bparser.py
```diff
@@ -12,6 +12,13 @@
 from bibtexparser import customization
 
 
+NONSTANDARD_ENTRY = """@nonstandard{mycitekey,
```
mwatts15: Why was this variable introduced? It doesn't seem to be used in this file.

omangin: My bad! Thanks @mwatts15.

Issue #198, opened 2018-03-22
Bibtexparser has issues with the preamble of the ACM bibtex file


lfdversluis:
The ACM includes some preamble in their example, which causes the following error when loading the file. Removing this preamble "fixes" it.

```
bib_database = bibtexparser.load(bibtex_file)
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/__init__.py", line 59, in load
    return parser.parse_file(bibtex_file)
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/bparser.py", line 130, in parse_file
    return self.parse(file.read())
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/bparser.py", line 119, in parse
    self._parse_records(customization=self.customization)
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/bparser.py", line 163, in _parse_records
    _add_parsed_record(record, records)
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/bparser.py", line 144, in _add_parsed_record
    parsed = self._parse_record(record, customization=customization)
  File "/home/lfdversluis/.local/lib/python2.7/site-packages/bibtexparser/bparser.py", line 227, in _parse_record
    key, val = [i.strip().strip('{').strip('}').replace('\n', ' ') for i in record.split('{', 1)[1].strip('}').strip('\n').strip(',').split('=')]
ValueError: too many values to unpack
```

Preamble:

```
%%% ====================================================================
%%%  BibTeX-file{
%%%     author          = "Gerry Murray",
%%%     version         = "1.2",
%%%     date            = "2 April 2012",
%%%     filename        = "acmsmall-sample-bibfile.bib",
%%%     address         = "ACM, NY",
%%%     email           = "murray at hq.acm.org",
%%%     codetable       = "ISO/ASCII",
%%%     keywords        = "ACM Reference Format, bibliography, citation, references",
%%%     supported       = "yes",
%%%     docstring       = "This BibTeX database file contains 'bibdata' entries
%%%                        that 'match' the examples provided in the Specifications Document
%%%                        AND, also, 'legacy'-type bibs. It should assist authors in 
%%%                        choosing the 'correct' at-bibtype and necessary bib-fields
%%%                        so as to obtain the appropriate ACM Reference Format output. 
%%%			   It also contains many 'Standard Abbreviations'. "
%%%  }
[...]

PR #199, opened 2018-03-22
Handles declarations on lines starting with spaces after comments.

omangin:
(See #198)

The parser currently skips declarations preceded by a space after a comment:
they are considered part of the implicit comment before. This behavior differs,
for example, from what biber would do on the same file. It is also not consistent
with the fact that a line started as " @string" is a valid string declaration.


sciunto:
Wonderful!

Issue #203, opened 2018-04-12
Handle undefined bibtex strings


omangin:
The idea is to be able to represent undefined bibtex strings as such even if the definition does not appear in the current bibdata. This would be an alternative to including and interpolating common strings and would enable exporting bibtex files identical to the input even when they include things such as `month = jan`.

noahdesu:
actually if @omangin or @sciunto could chime in on a design that'd be useful. something as simple as a "strict = False" flag on the parser could work, but we could also imagine something like passing in a function to resolve strings not in the database. That might be more general and just as easy to implement.

rwirth:
I've solved this by creating a BibtexParser instance with `interpolate_strings=False`. This leaves expressions involving strings as `BibDataStringExpression`s, which can be inspected or manually interpolated.

noahdesu:
@omangin it looks you had some provided some minor feedback on the #205 changes. Let me know if / when you decide to go with that approach and I'll address those comments.

omangin:
@sciunto Do you have an opinion on this. As a summary the idea is to be able to set a function that is used as a default to interpolate undefined strings. This has the advantage of handling undefined-but-common strings without having to actually add definitions into the bibdata's string dictionary.

As far as I am concerned I think it is more elegant then the current implementation and would be in favor of including it.

sciunto:
@omangin Thanks for raising this to my attention. Yes, I like this idea. So, +1.

alan-isaac:
Turning off string interpolation is helpful but not enough.  I'd like a warning but the undefined string unchanged (so that the .bib file is still successfully parsed).  Imo, a runtime crash on an undefined string seems like overkill: could you change `bibdatabase.expand_string` to warn instead of raise (at least until another solution is implemented)?

MiWeiss:
Note: #205 was close to a merge (tests & docs missing) but was otherwise decided to be acceptable.

However, it was closed by the PR opener. We'd need someone to clone that branch and complete it.

MiWeiss:
Stale PR no one is working on anymore. Seems not relevant for v2. Closing issue (#203) and PR (#205). If anyone wants to go back and look at the PR for a v1 fix, feel free.

PR #205, opened 2018-05-02
DNM: add a string provider interface

dotnwat:
#203 

Signed-off-by: Noah Watkins <noahwatkins@gmail.com>

### DIFF
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -48,7 +48,7 @@ class BibDatabase(object):
     Bibliographic database object that follows the data structure of a BibTeX file.
     """
 
-    def __init__(self):
+    def __init__(self, string_provider=None):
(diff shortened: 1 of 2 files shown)


bibtexparser/bibdatabase.py
```diff
@@ -101,6 +102,10 @@ def expand_string(self, name):
             return BibDataStringExpression.expand_if_expression(
                 self.strings[name])
         except KeyError:
+            if self.string_provider:
+                val = self.string_provider(name)
+                if val:
```
omangin: Here we want `val is not None` to allow empty string values. An alternative would be to expect the string provider to raise a `KeyError` or something similar. I think it is fine to allow both by making using the `is` as suggested.

sciunto:
Before merging this, we need:
* some unittests
* an additional section in the documentation. (eg in the pepper/salt section).

Thanks! :)

omangin:
Hi @noahdesu. Did you have any time to add address the comments above? It would be great to be able to merge this!

PR #208, opened 2018-07-02
Add support for BibLaTeX annotations

cao:
# Issue
BibLaTeX-style annotations result in entries being incorrectly parsed as `ImplicitComment`. 

For example, the following article entry will not parse because it contains "author+an":
```
@ARTICLE{Cesar2013,
  author = {Jean César},
  author+an = {1=highlight},
  title = {An amazing title},
  year = {2013},
  month = "jan",
  volume = {12},
  pages = {12-23},
  journal = {Nice Journal},
  abstract = {This is an abstract. This line should be long enough to test
multilines... and with a french érudit word},
  comments = {A comment},
  keyword = {keyword1, keyword2},
}
```

# Pull Request 
Support BibLaTeX-style annotations in entries by allowing the plus sign (+) in a field names. No support to use the annotations in any way is added.

# References
* [BibLaTeX Documentation](http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf), Section 3.6 Data Annotations (page 75)

### DIFF
--- a/bibtexparser/bibtexexpression.py
+++ b/bibtexparser/bibtexexpression.py
@@ -154,7 +154,7 @@ class BibtexExpression(object):
         key.setParseAction(lambda s, l, t: first_token(s, l, t).strip())
 
         # Field name: word of letters, digits, dashes and underscores
-        field_name = pp.Word(pp.alphanums + '_-().')('FieldName')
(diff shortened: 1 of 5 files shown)


cao:
Added a test for `bwriter`.

Issue #212, opened 2018-08-06
Possible exceptions are not documented


benureau:
On a parsing problem, most of the time, the user will receive either a `pyparsing.ParseException` or a `pyparsing.ParseSyntaxException`. But when providing the empty string, an `IndexError` will be raised. This is not a problem per se, but currently, this is not documented, and must be discovered when problem happen.

benureau:
And the following bibstring will throw a `bibtexparser.bibdatabase.UndefinedString`:
```
@misc{this looks,
  like = a = bibtex file but
    , is not a real one!
```

```python
bibraw = '''@misc{this looks,
  like = a = bibtex file but
    , is not a real one!
'''

from bibtexparser.bparser import BibTexParser

btp = BibTexParser()

print(btp.parse(bibraw, partial=False).get_entry_dict())
```
produces:

```
Traceback (most recent call last):
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/bibtexparser/bibdatabase.py", line 102, in expand_string
    self.strings[name])
KeyError: 'a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "pubs/test.py", line 23, in <module>
    print(btp.parse(bibraw, partial=False).get_entry_dict())
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/bibtexparser/bparser.py", line 147, in parse
    self._expr.parseFile(bibtex_file_obj)
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/bibtexparser/bibtexexpression.py", line 263, in parseFile
    return self.main_expression.parseFile(file_obj, parseAll=True)
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/pyparsing.py", line 2163, in parseFile
    return self.parseString(file_contents, parseAll)
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/pyparsing.py", line 1622, in parseString
    loc, tokens = self._parse( instring, 0 )
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/pyparsing.py", line 1379, in _parseNoCache
    loc,tokens = self.parseImpl( instring, preloc, doActions )
  File "/Users/fcyb/.venvs/pubs37/lib/python3.7/site-packages/pyparsing.py", line 3919, in parseImpl
[...]

PR #214, opened 2018-08-20
Fixes for #212

omangin:
- [X] Prevents failures on short strings with no BOM.
- [x] Add tests for short strings with no BOM.
- [x] Check string definitions after the full value is checked (otherwise may fail while pyparsing is probing whether the current chunk is formatted as a value).


bibtexparser/bibtexexpression.py
```diff
@@ -95,6 +95,8 @@ class BibtexExpression(object):
     ParseException = pp.ParseException
 
     def __init__(self):
+        # Init parse action functions
```
sciunto: Specifying the meaning of s, l, t in a comment somewhere would be good.

bibtexparser/bparser.py
```diff
@@ -212,12 +204,13 @@ def _init_expressions(self):
     def _bibtex_file_obj(self, bibtex_str):
         # Some files have Byte-order marks inserted at the start
         byte = b'\xef\xbb\xbf'
+
         if isinstance(bibtex_str, str):
-            byte = str(byte, self.encoding, 'ignore')
-            if bibtex_str[0] == byte:
+            byte = ustr(byte, self.encoding, 'ignore')
```
sciunto, line 209:
```suggestion
            byte = str(byte, self.encoding, 'ignore')
```

PR #216, opened 2018-08-24
Feature: crossref Support

cao:
# crossref Support

bibtexparser does currently not resolve cross references. This pull request adds the support.

It is based on #109 by @flopraden, credit for the feature goes to him.

This pull request deprecates #109 and expands on it:
- Rebased onto master
- Commit message style fixed
- Failing unit tests fixed
- Clarified BibTeX assumptions
- Typos and minor fixes 

# References
* [BibLaTeX Documentation](http://mirrors.ibiblio.org/CTAN/macros/latex/contrib/biblatex/doc/biblatex.pdf), Section 2.2.3 Special Fields (page 26)

### DIFF
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -1,9 +1,11 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 from collections import OrderedDict
 import sys
-
 import logging
-logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
 
 if sys.version_info.major == 2:
     TEXT_TYPE = unicode
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -114,14 +114,14 @@ class BibDatabase(object):
     def _add_missing_field_from_crossref_entry(self, entry, dependance = []):
         if entry["ID"] in self._crossref_updated:
             return
-        if entry["crossref"] not in self._entries_dict:
-            logger.error("Crossref reference %s for %s is missing.", entry["crossref"], entry["ID"])
+        if entry["_crossref"] not in self._entries_dict:
+            logger.error("Crossref reference %s for %s is missing.", entry["_crossref"], entry["ID"])
             return
-        if entry["crossref"] in dependance:
+        if entry["_crossref"] in dependance:
             logger.error("Circular crossref dependance : %s.", "->".join())
             return
-        crossref_entry = self._entries_dict[entry["crossref"]]
-        if "crossref" in crossref_entry:
+        crossref_entry = self._entries_dict[entry["_crossref"]]
+        if "_crossref" in crossref_entry:
             # update cross-ref for the cross-referenced entry
             dependance.append(self.entry["ID"])
             self._add_missing_field_from_crossref_entry(crossref_entry, dependance)
@@ -132,12 +132,13 @@ class BibDatabase(object):
         for bibfield,bibvalue in missing_field:
             entry[bibfield] = bibvalue
         self._crossref_updated.append(entry["ID"])
+        del entry["_crossref"]
 
     def _add_missing_field_from_crossref(self):
         self._make_entries_dict()
         self._crossref_updated = []
         for entry in self.entries:
-            if "crossref" in entry:
+            if "_crossref" in entry:
                 self._add_missing_field_from_crossref_entry(entry)
 
 
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -1,6 +1,9 @@
 from collections import OrderedDict
 import sys
 
+import logging
+logger = logging.getLogger(__name__)
+
 
 if sys.version_info.major == 2:
     TEXT_TYPE = unicode
@@ -84,14 +87,17 @@ class BibDatabase(object):
             result.append(TEXT_TYPE(entry.get(field, '')).lower())  # Sorting always as string
         return tuple(result)
 
+    def _make_entries_dict(self):
+        for entry in self.entries:
+            self._entries_dict[entry['ID']] = entry
+
     def get_entry_dict(self):
         """Return a dictionary of BibTeX entries.
         The dict key is the BibTeX entry key
         """
         # If the hash has never been made, make it
         if not self._entries_dict:
-            for entry in self.entries:
-                self._entries_dict[entry['ID']] = entry
+            self._make_entries_dict()
         return self._entries_dict
 
     entries_dict = property(get_entry_dict)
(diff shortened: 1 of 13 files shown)


cao:
Some unit tests are not yet deterministic (circular checking), and others rely on Python 3.4 (`assertLogs`). I'll be pushing a fix in a bit.

sciunto:
I triggered them. Thanks!

cao:
Thanks, looks like I was too optimistic to assume that Travis CI would support Python 3.7 by now.

Since it is unrelated to the crossref feature and would have sneaked in with the PR (so to speak), I reset HEAD to the previous commit and opened issue #217 to track the CI environment.

This PR is ready.

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
...
+                            bibfield not in self.not_update_by_crossref}
+
+        for bibfield, bibvalue in from_crossref.items():
+            entry[bibfield] = bibvalue
+
+        self._crossref_updated.append(entry['ID'])
+        entry['_FROM_CROSSREF'] = sorted(from_crossref.keys())
+        del entry['_crossref']
+
+    def _add_missing_from_crossref(self):
```
omangin: This should not be marked private since it is used in bparser.

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
...
+        self._make_entries_dict()
```
omangin: Indeed `_add_missing_from_crossref_entry` assumes that `_make_entries_dict` has been called, which should be documented in the former. As an alternative, `_add_missing_from_crossref_entry` could call `get_entry_dict()` instead of directly accessing the attribute, which would makes the code slightly less efficient but slightly less error-prone.

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
         except KeyError:
             raise(UndefinedString(name))
 
+    def _add_missing_from_crossref_entry(self, entry, dependencies=set()):
+        if entry['ID'] in self._crossref_updated:
+            return
+
+        if entry['_crossref'] not in self._entries_dict:
```
omangin: See comment in `_add_missing_from_crossref` about using the attribute `_entries_dict` here.

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
...
+            logger.error("Crossref reference %s for %s is missing.",
+                         entry['_crossref'],
+                         entry['ID'])
+            return
```
omangin: Don't we want to fail here?

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
...
+
+        if entry['_crossref'] in dependencies:
+            logger.error("Circular crossref dependency: %s->%s->%s.",
+                         "->".join(dependencies),
+                         entry['ID'],
+                         entry['_crossref'])
+            return
```
omangin: Same thing: we might want to just fail.

bibtexparser/bibdatabase.py
```diff
@@ -103,6 +116,48 @@ def expand_string(self, name):
...
+            self._add_missing_from_crossref_entry(crossref_entry, dependencies)
+            dependencies.remove(entry['ID'])
+
+        from_crossref = {bibfield: bibvalue
+                         for (bibfield, bibvalue) in crossref_entry.items()
+                         if bibfield not in entry.keys() and
+                            bibfield not in self.not_update_by_crossref}
+
+        for bibfield, bibvalue in from_crossref.items():
+            entry[bibfield] = bibvalue
```
omangin: `entry.update(from_crossref)`

bibtexparser/bibdatabase.py
```diff
@@ -64,6 +70,10 @@ def __init__(self):
         #: List of BibTeX preamble (`@preamble{...}`) blocks.
         self.preambles = []
 
+        self.not_update_by_crossref = [
```
omangin: `not_updated_by_crossref`

This also deserves some comment to explain its purpose.

cao:
@omangin Thanks for the comments, I agree with them. I won't be able to address them until late next week though. Please let me know if this would mean it misses the next release (are releases tracked somewhere?), because it would be great to have crossref support included in it.

omangin:
I believe next week will be fine. I don't think we have a very regular schedule for releases. As far as I understood they go when the code reaches a stable state with new features. But @sciunto will probably better answer this question.

sciunto:
Such additions can trigger a release without any problem. I'm not much involved in the development of this library, but I can do that job quite quickly :)

About the question on exceptions vs logging. I would vote for logging. My arguments:
The job is to parse, not to check extensively the consistency. Of course, if parsing is not possible due to mistakes, we have to raise an exception. If the lib notices something weird, we can advise with the logging module, but the user should not have the feeling to buy a guarantee. bibtex is already a nightmare :)

bibtexparser/bibdatabase.py
cao: Fixed in 66144482d40b819abab87f774895211035dadd4d.

bibtexparser/bibdatabase.py
cao: Fixed in e0ab2c257bfb9853855b955b845a14aa7d7ee7d3.

bibtexparser/bibdatabase.py
cao: I think logging would be more reasonable then failing because the entry is valid, and it only does not resolve an unnecessary crossref properly. If we visit the same entry through a crossref chain again, then we already updated every field of the current entry with it (at the earlier visit), so we can't get anything "new." In turn, we can just stop here, which is what we do.

bibtexparser/bibdatabase.py
cao: I don't think it makes sense to fail here, because biber/biblatex etc. are also not failing but logging an undefined reference instead, and consistency with them is probably what users expect the behavior to be.

bibtexparser/bibdatabase.py
cao: Fixed in 74ae491ca8af740458f90ac049d25c2a17c49a95.

bibtexparser/bibdatabase.py
cao: Fixed in 74ae491ca8af740458f90ac049d25c2a17c49a95.

bibtexparser/bibdatabase.py
cao: Fixed in 3139f82ac4029034b0242dcae6b2638fb63fc9d5.

bibtexparser/bibdatabase.py
cao: I marking the conversation as resolved, so that GitHub updates the PR correctly.

bibtexparser/bibdatabase.py
cao: Marking as resolved, so that GitHub updates PR properly.

omangin:
Perfect! Thanks again @cao and @flopraden.

Please fill free to add a quick section in the tutorial to explain how this should be used.

PR #207, opened 2018-05-07
Add comma_last to BibTexWriter.

u55:
Closes #206.

PR #219, opened 2018-09-07
New add_trailing_comma option in BibTexWriter. Closes #206, with tests.

### LINKED ISSUE #206: ENH: Optional last comma in BibTexWriter
BibTeX allows the comma to be optional at the end of the last field in an entry. Currently, `BibTexWriter` only writes entries with this comma missing, but I would like to have the option of including it.  For example, both of these entries are valid:
```
@ARTICLE{Cesar2013,
  author = {Jean Cesar},
  title = {An amazing title}
}
@ARTICLE{Cesar2013,
  author = {Jean Cesar},
  title = {An amazing title},
}
```
Including the ending comma has the advantage that if I later manually add a new field at the end, I don't have to modify the previous line. This also makes for cleaner file diffs.

I am thinking of something as simple as adding `self.comma_last=False` to `BibTexWriter.__init__`, and modifying [the ending line of `BibTexWriter._entry_to_bibtex`](https://github.com/sciunto-org/python-bibtexparser/blob/f5b1bf8cceeba88367c9e1598133df83b60d80c7/bibtexparser/bwriter.py#L139) to be:

```python
if self.comma_last:
    bibtex += ",\n}\n" + self.entry_separator
else:
    bibtex += "\n}\n" + self.entry_separator
```
Thanks.

u55:


u55:
This is a rebased replacement for pull request #207 which closes issue #206.

PR #234, opened 2019-02-17
common strings where not present on parsing for months

agramfort:
I need this "hack" to have the months properly recognized.

### DIFF
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -72,6 +72,7 @@ class BibDatabase(object):
 
         #: List of fields that should not be updated when resolving crossrefs
         self._not_updated_by_crossref = ['_FROM_CROSSREF']
+        self.load_common_strings()
 
     def load_common_strings(self):
         self.strings.update(COMMON_STRINGS)


omangin:
Hi @agramfort. Thank you for your contribution. Could you be more explicit on the situation where this hack is needed?

It could be that what you need is to set the `common_strings=true` in the parser constructor: this is the place where it is handled currently. If you are directly using the `Database` class you can call the method manually. (See also the note in [this tutorial section](https://bibtexparser.readthedocs.io/en/master/tutorial.html#simplest-call).)

agramfort:
I did this:

```
import bibtexparser

with open('./data/Gramfort.bib') as bib:
    bib_str = bib.read()

records = bibtexparser.loads(bib_str)
```

file is here https://github.com/agramfort/alexandre.gramfort.net/blob/master/data/Gramfort.bib

I started to see the problem when updating bibtexparser (that I had not updated in > 2 years)

omangin:
Thanks. And, yes, the API changed in the last two years! Your bibtex file contains attributes like `month = Jun` where the right part is not a regular bibtex value but assumes that a bibtex string named `Jun` is defined. This is not purely bibtex but is common, so it is now not handled by default but if you pass the option `common_string` to the parser constructor. As mentioned in the tutorial linked above, the `load` function that you are using hides the creation of the parser so you have to be explicit instead and use instead:
```python
bib_database = bibtexparser.bparser.BibTexParser(common_strings=True).parse(bibtex_str)
```

Issue #195, opened 2018-03-09
Option to exclude fields from export/writer


ar-jan:
An option to omit specified fields from the BibTexWriter would be useful (unless I've missed some other way to achieve this, of course!). We could parse the bibtex database and remove fields, but I have a situation where I need a field to be in the bibtex database for sorting, but it should not be exported.

omangin:
Indeed, that would be a valuable option since I believe it is a common use case. It should also be pretty straightforward to implement.

MiWeiss:
Stale pull request for v1. Easily implementable for v2 through custom middleware (currently no builtin).

PR #250, opened 2020-02-05
Select fields for export with BibTexWriter class

andrejberg:
Now a list can be given of fields. Only those fields are returned in BibTexWriter is used to generate a string. This is not exactly what was asked in issue #195 but it should do the job.

### DIFF
--- a/bibtexparser/bwriter.py
+++ b/bibtexparser/bwriter.py
@@ -77,6 +77,9 @@ class BibTexWriter(object):
         self._max_field_width = 0
         #: Whether common strings are written
         self.common_strings = write_common_strings
+        #: Define which fields will be written out for entries (list)
+        #: default is empty list: all available fields are writen
+        self.export_fields = []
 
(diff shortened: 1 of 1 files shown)


MiWeiss:
Hi @andrejberg. 

I just joined the project and wanted to give you a quick heads up. 

In principle, I like the proposition. However, I am a bit worried that we're starting to overload the writer parameters. If we add this, it won't take long until we'll add an "exclude_fields" besides the here proposed "include_fields", and I see a range of other additions suggested in other issues. Maybe it would be more sustainable if we made this a bit bigger, and added the option to pass in customization functions, similar to the parsing.

I'll thus keep this open for just a bit longer, until a decision on the above was taken.

MiWeiss:
Close as stale: No reaction to previous comments. Also, workaround easily implementable by just modifying the BibTeX database entries directly or by writing a simple customizer on parsing.

PR #251, opened 2020-02-21
Use https://doi.org/ for DOIs

kno10:
New DOI url style.


sciunto:
Thank you, good to switch to https as well.

PR #255, opened 2020-03-25
Remove shebang lines from non-executable Python files

flyn-org:
I came across this while packaging python-bibtexparser for Fedora. See
https://bugzilla.redhat.com/show_bug.cgi?id=1812973

Signed-off-by: W. Michael Petullo <mike@flyn.org>


sciunto:
Thank you @flyn-org !

PR #277, opened 2021-02-13
Allow preservation of existing order of entry fields in writer

alexreg:
This should be fairly self-explanatory. I couldn't immediately see why `display_order = reversed(entry)` is correct to preserve the pre-existing order of fields, but it seems to work reliably.

### DIFF
--- a/bibtexparser/bwriter.py
+++ b/bibtexparser/bwriter.py
@@ -114,19 +114,20 @@ class BibTexWriter(object):
         return bibtex
 
     def _entry_to_bibtex(self, entry):
+        from io import StringIO
         from itertools import chain
 
-        bibtex = ''
+        bibtex = StringIO()
         # Write BibTeX key
-        bibtex += '@' + entry['ENTRYTYPE'] + '{' + entry['ID']
+        bibtex.write('@' + entry['ENTRYTYPE'] + '{' + entry['ID'])
 
         if self.display_order is not None:
             # create display_order of fields for this entry
             # first those keys which are both in self.display_order and in entry.keys
             # then all the other fields sorted alphabetically
             display_order = chain(
-                i for i in self.display_order if i in entry,
-                i for i in sorted(entry) if i not in self.display_order)
+                (i for i in self.display_order if i in entry),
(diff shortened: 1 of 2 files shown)


alexreg:
FYI I added a new test, so coveralls shouldn't be complaining now, though I think it forgot to rerun.

tdegeus:
Is this aimed at preserving 
```
@article{foo,
 title = {I would like to change as few as possible for version management},
 year = {20202}
 author = {Foo, B.},
}
```
(the order of `title`, `year`, `author` in this example).
That would be amazing for version management!

MiWeiss:
Hi @alexreg, and sorry for the wait :-)

I just started helping this repo a bit with maintenance and I'd like to get this merged. Are you still interested in working on this PR? If so, a few remarks:

You asked why reversing the dict gives the original order. I guess that's caused by [this line](https://github.com/sciunto-org/python-bibtexparser/blob/master/bibtexparser/bibtexexpression.py#L191), where the fields found in pyparsing are reversed. They are later stored in a dict, which, as of python3.7 are [guaranteed](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) to preserve order. 

A few comments on your PR, before going into a deeper review:
- Please resolve merge conflicts.
- Please move the tidying of the literals in the test into a separate PR. 
- Please document your changes in the corresponding docstring.
- Python versions pre-dating 3.7 reached EOL, hence I guess we can continue working with `dict` instead of `OrderedDict`. But we may want to raise a warning if your enhancement is called (i.e, `display_order` set to `None`) with previous python versions. What do you think?

If you're not interested in continuing with this, please give me a quick heads-up - I'll hope to find someone else, then :-)

alexreg:
@MiWeiss Thanks for your response. I'm afraid I don't have the time to rework this PR right now (and I haven't been using this library lately), so if you could find someone else, that would be great.

MiWeiss:
Thanks a lot for getting back with me so quickly!

michaelfruth:
I can take this feature 👍.

MiWeiss:
Replaced in #317 by @michaelfruth. Thanks 👍

PR #282, opened 2021-10-27
Author split case-insensitive

tdegeus:
I have recently parsed a BibTeX library of a colleague who used `AND` to separate authors on some occasions


MiWeiss:
Merged as part of #299. Thanks!

PR #295, opened 2022-05-21
Remove obsolete unittest2 support

mgorny:
Now that the package does not support Python 2 anymore, there is
no reason to ever use unittest2.


MiWeiss:
Merged as part of #299. Thanks!

PR #296, opened 2022-05-31
Update bibdatabase.py

gideonite:
Make use of the COMMON_STRINGS variable

COMMON_STRINGS currently contains string aliases of month names. Before this change, the parser fails when it encounters these aliases for the "month" field of a bibtex item.


MiWeiss:
Not sure if I understand you rPR correctly, but to use the common strings during parsing, do the following:

`bib_database` = bibtexparser.bparser.BibTexParser(common_strings=True).parse_file(bibtex_file)

I'm thinking about making `common_strings=True` default, but for now, the above is what's described in the [docs](https://bibtexparser.readthedocs.io/en/master/tutorial.html)

Issue #308, opened 2022-07-10
Warn if parser called multiple times


MiWeiss:
Consider the snippet shown below. The second `parses` call returns 4 entries, I would have expected three.

I do see use cases where this is intended, and changing this could break some dependent applications. Hence, I propose to raise a warning if the parse function is called more than once, with some way to disable the warning.

(issue identified as part of #247)

```
import bibtexparser

from bibtexparser.bparser import BibTexParser

parser = BibTexParser()

a = """
@article{Cesar2013,
  author = {Jean César},
  title = {An amazing title},
  year = {2013},
  month = "jan",
  volume = {12},
  pages = {12-23},
  journal = {Nice Journal},
  abstract = {This is an abstract. This line should be long enough to test
multilines... and with a french érudit word},
  comments = {A comment},
  keyword = {keyword1, keyword2},
}
"""
a_database = bibtexparser.loads(a, parser)

print(len(a_database.entries))

b = """
@article{FuMetalhalideperovskite2019,
    author = {Yongping Fu and Haiming Zhu and Jie Chen and Matthew P. Hautzinger and X.-Y. Zhu and Song Jin},
    doi = {10.1038/s41578-019-0080-9},
    journal = {Nature Reviews Materials},
    month = {feb},
    number = {3},
    pages = {169-188},
    publisher = {Springer Science and Business Media {LLC}},
    title = {Metal halide perovskite nanostructures for optoelectronic applications and the study of physical properties},
    url = {https://www.nature.com/articles/s41578-019-0080-9},
    volume = {4},
    year = {2019}
}

@article{SunEnablingSiliconSolar2014,
    author = {Ke Sun and Shaohua Shen and Yongqi Liang and Paul E. Burrows and Samuel S. Mao and Deli Wang},
    doi = {10.1021/cr300459q},
    journal = {Chemical Reviews},
    month = {aug},
    number = {17},
    pages = {8662-8719},
    publisher = {American Chemical Society ({ACS})},
    title = {Enabling Silicon for Solar-Fuel Production},
    url = {http://pubs.acs.org/doi/10.1021/cr300459q},
    volume = {114},
    year = {2014}
}

@article{LiuPhotocatalytichydrogenproduction2016,
    author = {Maochang Liu and Yubin Chen and Jinzhan Su and Jinwen Shi and Xixi Wang and Liejin Guo},
    doi = {10.1038/nenergy.2016.151},
    impactfactor = {54.000},
[...]

PR #312, opened 2022-07-10
:children_crossing: Raise warning if parser is used multiple times

mrn97:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/308. A warning message will appear if the `parse` method is called more than once and the message will be "Parser has been called more than once."

### DIFF
--- a/bibtexparser/bparser.py
+++ b/bibtexparser/bparser.py
@@ -76,8 +76,7 @@ class BibTexParser(object):
                  homogenize_fields=False,
                  interpolate_strings=True,
                  common_strings=True,
-                 add_missing_from_crossref=False,
-                 expect_multiple_parse=False):
+                 add_missing_from_crossref=False):
         """
         Creates a parser for reading BibTeX files
 
@@ -85,8 +84,8 @@ class BibTexParser(object):
         :rtype: `BibTexParser`
         """
 
-        self.expect_multiple_parse = expect_multiple_parse
         self._parse_call_count = 0
+        self.expect_multiple_parse = False
 
         self.bib_database = BibDatabase()
 
@@ -139,7 +138,6 @@ class BibTexParser(object):
 
     def parse(self, bibtex_str, partial=False):
         """Parse a BibTeX string into an object
-        :param expect_multiple_parse: if True, does not print warnings
         :param bibtex_str: BibTeX string
         :type: str or unicode
         :param partial: If True, print errors only on parsing failures.
@@ -150,9 +148,10 @@ class BibTexParser(object):
         """
 
         self._parse_call_count += 1
-        call_counter = self._parse_call_count
-        if call_counter > 1 and not self.expect_multiple_parse:
-            warnings.warn("parser has been called more than once")
+        if self._parse_call_count == 2 and not self.expect_multiple_parse:
+            warnings.warn("Parser has been called more than once, "
+                          "avoid the warning by setting property expect_multiple_parse to True.", category=UserWarning,
+                          stacklevel=2)
 
         bibtex_file_obj = self._bibtex_file_obj(bibtex_str)
         try:
@@ -189,7 +188,7 @@ class BibTexParser(object):
         # Handle string as BibDataString object
         self._expr.set_string_name_parse_action(
             lambda s, l, t:
-                BibDataString(self.bib_database, t[0]))
+            BibDataString(self.bib_database, t[0]))
 
         # Add notice to logger
         self._expr.add_log_function(logger.debug)
@@ -198,20 +197,20 @@ class BibTexParser(object):
         self._expr.entry.addParseAction(
             lambda s, l, t: self._add_entry(
                 t.get('EntryType'), t.get('Key'), t.get('Fields'))
-            )
+        )
         self._expr.implicit_comment.addParseAction(
             lambda s, l, t: self._add_comment(t[0])
-            )
+        )
         self._expr.explicit_comment.addParseAction(
             lambda s, l, t: self._add_comment(t[0])
-            )
+        )
         self._expr.preamble_decl.addParseAction(
             lambda s, l, t: self._add_preamble(t[0])
-            )
+        )
         self._expr.string_def.addParseAction(
             lambda s, l, t: self._add_string(t['StringName'].name,
                                              t['StringValue'])
-            )
+        )
 
     def _bibtex_file_obj(self, bibtex_str):
         # Some files have Byte-order marks inserted at the start
--- a/bibtexparser/bparser.py
+++ b/bibtexparser/bparser.py
@@ -150,9 +150,10 @@ class BibTexParser(object):
 
         self._parse_call_count += 1
         if self._parse_call_count == 2 and not self.expect_multiple_parse:
-            warnings.warn("Parser has been called more than once, "
-                          "avoid the warning by setting property expect_multiple_parse to True.", category=UserWarning,
-                          stacklevel=2)
+            warnings.warn("The parser has been called more than once. "
+                          "Subsequent parse calls lead to a combined BibTeX library. \n"
+                          "To avoid the warning, set property `parser.expect_multiple_parse` to `True`.",
+                          category=UserWarning, stacklevel=2)
 
         bibtex_file_obj = self._bibtex_file_obj(bibtex_str)
         try:
--- a/bibtexparser/bparser.py
+++ b/bibtexparser/bparser.py
@@ -85,6 +85,8 @@ class BibTexParser(object):
         """
 
         self._parse_call_count = 0
+        #: Parsers can be called multiple times, merging databases, but raising a warning.
+        #      Set this to `True` to disable the warning. Default: `False`
         self.expect_multiple_parse = False
 
         self.bib_database = BibDatabase()
--- a/bibtexparser/bparser.py
+++ b/bibtexparser/bparser.py
@@ -138,6 +138,7 @@ class BibTexParser(object):
 
     def parse(self, bibtex_str, partial=False):
         """Parse a BibTeX string into an object
+
         :param bibtex_str: BibTeX string
         :type: str or unicode
         :param partial: If True, print errors only on parsing failures.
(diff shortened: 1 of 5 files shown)


bibtexparser/bparser.py
```diff
@@ -75,13 +76,18 @@ def __init__(self, data=None,
                  homogenize_fields=False,
                  interpolate_strings=True,
                  common_strings=True,
-                 add_missing_from_crossref=False):
+                 add_missing_from_crossref=False,
+                 expect_multiple_parse=False):
```
MiWeiss, line 80: I suggest to *not* make this a constructor argument (instead users should modify the property).

Reasons: Keeping the constructor less verbose by not including options which will hardly ever be used.

bibtexparser/bparser.py
```diff
@@ -133,7 +139,7 @@ def __init__(self, data=None,
 
     def parse(self, bibtex_str, partial=False):
         """Parse a BibTeX string into an object
-
+        :param expect_multiple_parse: if True, does not print warnings
```
MiWeiss, line 142: Remove this line

bibtexparser/bparser.py
```diff
@@ -142,6 +148,12 @@ def parse(self, bibtex_str, partial=False):
         :return: bibliographic database
         :rtype: BibDatabase
         """
+
+        self._parse_call_count += 1
+        call_counter = self._parse_call_count
+        if call_counter > 1 and not self.expect_multiple_parse:
+            warnings.warn("parser has been called more than once")
```
MiWeiss, line 155: Improve the printed warning: 
- Choose the warning level as one above
- UserWarning type
- Explain how to avoid the warning (setting property `expect_multiple_parse` to true)
- Fix capitaliation

bibtexparser/bparser.py
```diff
@@ -142,6 +148,12 @@ def parse(self, bibtex_str, partial=False):
         :return: bibliographic database
         :rtype: BibDatabase
         """
+
+        self._parse_call_count += 1
+        call_counter = self._parse_call_count
+        if call_counter > 1 and not self.expect_multiple_parse:
```
MiWeiss, line 154: `call_counter==1`, such that the warning is raised only once per parser.

MiWeiss (changes requested):
Thanks for your PR.
I remarked some required changes inline.

Please also add tests that cover the functionality, e.g.:
- Parsing just once should *not* raise your warning with default settings.
- Parsing twice should raise your warning with default settings.
- Parsing three times should raise warnings only once with default settings.
- Parsing three times should *not* raise a warning if `expect_multiple_parse=True`

bibtexparser/bparser.py
mrn97:
```
self._parse_call_count += 1
call_counter = self._parse_call_count
if call_counter == 1 and not self.expect_multiple_parse:
    warnings.warn("Parser has been called more than once, "
                  "avoid the warning by setting property expect_multiple_parse to true", category=UserWarning,
                  stacklevel=2)
```

Would it be ok?

bibtexparser/bparser.py
mrn97: What do you mean by the property? property with using decorators like `@property` and etc. or an instance attribute like below code would be ok? and if I remove it from the constructor how can I pass it as a flag?

```
# Removing from the constructor arguments 
self.expect_multiple_parse = False
```

Usage:

```
parser = BibTexParser()
parser.expect_multiple_parse = False
```

bibtexparser/bparser.py
MiWeiss: Yes, you are of course right. That was a typo, and the equality should be `==2`

bibtexparser/bparser.py
MiWeiss: Have not tested it yet, but yes, that looks about right. 👍

bibtexparser/bparser.py
MiWeiss: For now, your proposition is fine (that's how the remainder of this library, which dates many python versions back, is set up. in addition, if we ever need to change to `@property`, we will be able to do so without breaking backward compatibility.

bibtexparser/bparser.py
MiWeiss: One point I forgot to mention previously: This assignment to a local variable appears not to be needed: `call_counter = self._parse_call_count`. Just use `self._parse_call_count` in the comparison...

MiWeiss:
Excuse my late reply. 

On some files, you changed the indentation of the lines, which makes this PR a bit verbose. Would you mind reverting these?

mrn97:
I fixed the indentation as it was before.

bibtexparser/tests/test_bparser.py
```diff
@@ -85,18 +87,18 @@ def test_article(self):
...
-                              'journal': 'Nice Journal',
-                              'ID': 'Cesar2013',
-                              'pages': '12-23',
-                              'title': 'An amazing title',
-                              'comments': 'A comment',
-                              'author': 'Jean César',
-                              'volume': '12',
-                              'month': 'jan'
-                              }}
+                                           'ENTRYTYPE': 'article',
```
MiWeiss, line 90: Please revert these indentation changes.

bibtexparser/tests/test_bparser.py
```diff
@@ -337,7 +339,7 @@ def test_article_no_braces(self):
                      'title': 'An amazing title',
                      'abstract': "This is an abstract. This line should be long enough to test\nmultilines... and with a french érudit word",
                      },
-                     ]
+                    ]
```
MiWeiss, line 342: Please revert these spacing changes (also in the following files)

bibtexparser/tests/test_bparser.py
```diff
@@ -562,10 +564,10 @@ def test_string_definitions(self):
         res = dict(bib.strings)
         expected = COMMON_STRINGS.copy()
         expected.update({
-                'nice_journal': 'Nice Journal',
```
MiWeiss, line 565: Same here

bibtexparser/tests/test_bparser.py
```diff
@@ -607,8 +609,8 @@ def test_string_is_not_interpolated(self):
 
     def test_comments_spaces_and_declarations(self):
         with codecs.open(
-                'bibtexparser/tests/data/comments_spaces_and_declarations.bib',
```
MiWeiss, line 610: And here....

MiWeiss (changes requested):
> Sorry about that, I just fixed the indentation as it was before.

There still seem to be a bunch of them missing. I highlighted some (but not all) of them in the review above. You may want to check your IDE - maybe it is doing some autoformatting when comitting?


FYI: I do not dislike the changed indentations, but the PR should be self-contained and only consist of the changes relevant to the new feature (the warning), to keep the commit history as legible as possible.

mrn97:
> FYI: I do not dislike the changed indentations, but the PR should be self-contained and only consist of the changes relevant to the new feature (the warning), to keep the commit history as legible as possible.

I understand, you are right, it's now much cleaner and less verbose, I had to use `diff` from the beginning, I double-checked them using `git diff master issue308 file.py` and I think that's ok, but if you see any other errors I can fix it.

MiWeiss:
Thanks :-)

We're almost done. Unfortunately, I had previously not noticed that the warning provides relatively little information. In my review, I suggested a slightly different warning. Feel free to change my suggestion, however :-)

Afterward, if nothing else pops up, we should be ready to merge 🎉

mrn97:
>the warning provides relatively little information. In my review, I suggested a slightly different warning.

If I'm not mistaken you want `Traceback` to provide more information correct?

Is it ok with you?
```
  File "/testing/reproduce_issue247.py", line 15, in <module>
    b_database = bibtexparser.load(fileb, parser)
  File "/python-bibtexparser/venv/lib/python3.10/site-packages/bibtexparser-1.3.0-py3.10.egg/bibtexparser/__init__.py", line 69, in load
  File "/python-bibtexparser/venv/lib/python3.10/site-packages/bibtexparser-1.3.0-py3.10.egg/bibtexparser/bparser.py", line 181, in parse_file
  File "/python-bibtexparser/venv/lib/python3.10/site-packages/bibtexparser-1.3.0-py3.10.egg/bibtexparser/bparser.py", line 154, in parse
UserWarning: Parser has been called more than once, avoid the warning by setting property expect_multiple_parse to True.
```

MiWeiss:
> If I'm not mistaken you want `Traceback` to provide more information correct?

No, this is not what I meant ;-)

What I meant was that the current message in the warning "Parser has been called more than once, avoid the warning by setting property expect_multiple_parse to True." does not really give the user a good idea about why this may be a problem. 

I am thus suggesting to use the following instead:

> Parser has been called more than once. Subsequent parse call lead to a combined BibTeX library.
> To avoid the warning, set property `parser.expect_multiple_parse` to `True`.

You should see a "suggestion" in your review, on which - if you agree and don't spot any typos - can just click "accept" (or something alike):
image

MiWeiss (changes requested):

mrn97:
How can I revert the last commit because I added a suggestion?

bibtexparser/bparser.py
```diff
@@ -82,6 +83,10 @@ def __init__(self, data=None,
         :return: parser
         :rtype: `BibTexParser`
         """
+
+        self._parse_call_count = 0
+        self.expect_multiple_parse = False
```
MiWeiss, line 88:
```suggestion
        #: Parsers can be called multiple times, merging databases, but raising a warning. Set this to `True` to disable the warning.
        self.expect_multiple_parse = False
```

MiWeiss:
Hi @mrn97 

I am not sure why you'd want to revert a commit?
I have still added some missing documentation - not this PR looks good to me, and is ready to be merged. Any objections?

MiWeiss (approved):

MiWeiss:
@mrn97 is there anything you want to add to this? otherwise I'm gonna merge this soon

MiWeiss:
This is now merged. Thanks @mrn97 for your contribution

mrn97:
Excuse my late reply.
@MiWeiss Thank you for giving me the opportunity to contribute and for your help.

PR #316, opened 2022-09-05
:bug: align_values allows int or bool, fixes previous min alignment

### LINKED ISSUE #315: `align_values` has minimum `_max_field_width ` due to `ENTRYPOINT` pseudo field.
When using `align_values`, the minimum `_max_field_width` is always 9 as `ENTRYPOINT` (9 chars) is in the list of `entries`:
https://github.com/sciunto-org/python-bibtexparser/blob/006f463a8601dc8065f02f3199210c70f24ac4f9/bibtexparser/bwriter.py#L108

Therefore, this tests works fine:

```
    def test_align2(self):
        bib_database = BibDatabase()
        bib_database.entries = [{'ID': 'abc123',
                                 'ENTRYTYPE': 'book',
                                 'abc': 'test'}]
        writer = BibTexWriter()
        writer.align_values = True
        result = bibtexparser.dumps(bib_database, writer)
        expected = \
"""@book{abc123,
 abc       = {test}
}
"""
        self.assertEqual(result, expected)
```

Shouldn't the result be:

```
    def test_align2(self):
        bib_database = BibDatabase()
        bib_database.entries = [{'ID': 'abc123',
                                 'ENTRYTYPE': 'book',
                                 'abc': 'test'}]
        writer = BibTexWriter()
        writer.align_values = True
        result = bibtexparser.dumps(bib_database, writer)
        expected = \
"""@book{abc123,
 abc = {test}
}
"""
        self.assertEqual(result, expected)
```

This is also the behavior when having multiple short named keys.

michaelfruth:
Fixes #315

Aligns all values according to a given length by padding with single spaces. If align_values is true, the maximum number of characters used in any field name is used as the length. If align_values is a number, the greater of the specified value or the maximum number of characters used in any field name is used as the length.

> **_NOTE_:**: This commit includes breaking changes: The `ENTRYTYPE` entry was considered for calculating the maximum number of characters for entries, which always lead to a minimum value of `9`. Now, keys that are not written as BibTeX output are ignored, which leads to an exact computation of the number of characters for entries.

### DIFF
--- a/bibtexparser/bwriter.py
+++ b/bibtexparser/bwriter.py
@@ -94,8 +94,8 @@ class BibTexWriter(object):
         self.indent = ' '
         #: Align values. Aligns all values according to a given length by padding with single spaces.
         #    If align_values is true, the maximum number of characters used in any field name is used as the length.
-        #    If align_values is a number, the greater of the specified value or the maximum number of characters used
-        #    in any field name is used as the length.
+        #    If align_values is a number, the greater of the specified value or the number of characters used in the
+        #    field name is used as the length.
         #    Default: False
         self.align_values: Union[int, bool] = False
         #: Align multi-line values. Formats a multi-line value such that the text is aligned exactly
@@ -149,13 +149,13 @@ class BibTexWriter(object):
         else:
             entries = bib_database.entries
 
-        if self.align_values is True or type(self.align_values) == int:
+        if self.align_values is True:
             # determine maximum field width to be used
             widths = [len(ele) for entry in entries for ele in entry if ele not in ENTRY_TO_BIBTEX_IGNORE_ENTRIES]
-            if type(self.align_values) == int:
-                # specified width should be taken into account
-                widths.append(self.align_values)
             self._max_field_width = max(widths)
+        elif type(self.align_values) == int:
+            # Use specified value
+            self._max_field_width = self.align_values
 
         return self.entry_separator.join(self._entry_to_bibtex(entry) for entry in entries)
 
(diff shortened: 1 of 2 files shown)


MiWeiss:
Hi @michaelfruth 

I have not forgotten this one - and I have also noticed your comments in the other PRs  :-)

Regarding this PR: There's an edge case I still need to test & lay out to you. 
Regarding the other PRs: There's some longer thoughts I have organize in my head and then write down... :-)

I'm quite busy atm, but hope that I can get back with you ASAP. Sorry for the wait - and thanks for your contribution!

MiWeiss:
Hi @michaelfruth 

Excuse my late answer and thanks again for your PR. 

There's an edge case of which I am not sure how it should be handled: Assume someone specifies an `align_value=x` with int `x`, but there is some key `len(key) > x`.  As I understand it, the current implementation uses `len(x)`, ignoring the user-passed value.  As an alternative implementation, I could see that `x` is applied where possible, and only the too long key "exceeds it", e.g.

```
@entry{someentry,
lor    = "lorem ipsum",
lo     = "lorem ipsum",
lore   = "lorem ipsum",
loremipsum = "lorem ipsum",
l      = "lorem ipsum",
}
```

I guess this would be more closely to what I'd expect given the naming `align_value`. Also, it would probably be more useful for large to huge bibtex files, where a single key (like some unexpected value) should not mess up the formatting of the entire output. 

What do you think? Any reasons to prefer the current implementation?

MiWeiss:
@michaelfruth - I guess we'll be able to take this over to v2, hence I recommend we continue with this PR (also merging it into 1.x)

MiWeiss:
Hi @michaelfruth 

Not sure if this (again) ready for review. But in case it is - could you fix the merge conflict first? Would allow to run the actions :-)

michaelfruth:
Now PR is ready for review again 👍

MiWeiss (approved):

PR #317, opened 2022-09-05
✨Allow preservation of existing order of entry fields in writer

michaelfruth:
Fixes #277

Add different sorting strategies for elements that are added at the end of `display_order`. I'm not sure if we should use a `enum` for `display_order_sorting` or any other global value.. A enum is most convenient I think.

### DIFF
--- a/bibtexparser/bwriter.py
+++ b/bibtexparser/bwriter.py
@@ -22,7 +22,7 @@ class SortingStrategy(Enum):
     PRESERVE = auto()
 
 
-def apply_sorting_strategy(strategy: SortingStrategy, items: Iterable[str]) -> Iterable[str]:
+def _apply_sorting_strategy(strategy: SortingStrategy, items: Iterable[str]) -> Iterable[str]:
     if strategy == SortingStrategy.ALPHABETICAL_ASC:
         return sorted(items)
     elif strategy == SortingStrategy.ALPHABETICAL_DESC:
@@ -144,7 +144,7 @@ class BibTexWriter(object):
         # first those keys which are both in self.display_order and in entry.keys
         display_order = [i for i in self.display_order if i in entry]
         # then all the other fields sorted alphabetically
-        display_order += [i for i in apply_sorting_strategy(self.display_order_sorting, entry) if i not in self.display_order]
+        display_order += [i for i in _apply_sorting_strategy(self.display_order_sorting, entry) if i not in self.display_order]
         if self.comma_first:
             field_fmt = u"\n{indent}, {field:<{field_max_w}} = {value}"
         else:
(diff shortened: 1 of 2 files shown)


MiWeiss:
Thanks @michaelfruth for your PRs. Much appreciated.

I'm currently quite busy, and it will probably take a couple of days until I get to submit my reviews(*). Feel free to bump me if you think it's taking to long ;-)


(*) After a short glimpse, I do not expect any major comments...

bibtexparser/bwriter.py
```diff
@@ -14,6 +16,23 @@
 __all__ = ['BibTexWriter']
 
 
+class SortingStrategy(Enum):
+    ALPHABETICAL_ASC = auto()
+    ALPHABETICAL_DESC = auto()
+    PRESERVE = auto()
+
+
+def apply_sorting_strategy(strategy: SortingStrategy, items: Iterable[str]) -> Iterable[str]:
```
MiWeiss, line 25: I guess this method can be private (i.e., have a leading `_`)?

bibtexparser/bwriter.py
```diff
@@ -14,6 +16,23 @@
 __all__ = ['BibTexWriter']
 
 
+class SortingStrategy(Enum):
```
MiWeiss, line 19: I think we need some docstrings on this enum. There seems to be no universal agreement on how to do that - I'd propose to go along with the following: https://stackoverflow.com/a/72775639

MiWeiss (changes requested):
Thanks a lot for the PR, @michaelfruth. There's only two minor things I noticed. 

I have not yet manually tested the new feature, but given the unit tests, I expect no surprises :-)

bibtexparser/bwriter.py
michaelfruth: 👍 . I've added a docstring, but I'm not sure about whether this docstring will be processed properly: 
```
:py:attr:`~.BibTexWriter.display_order`
```

PyCharm doesn't render it..

bibtexparser/bwriter.py
michaelfruth: True. Changed scope of method.

bibtexparser/bwriter.py
MiWeiss: Yeh. Enum/Attribute docstrings are quite a hassle. And it's not getting better when you combine them. 

Either way, I think the current version is as good as we can do. Should future python versions define some better syntax, we can quickly change it. Thanks!

MiWeiss (approved):

MiWeiss:
Thanks @michaelfruth :-)

Issue #329, opened 2022-10-09
Remove "Not yet implemented" messages from seemingly implemented features from the docs


CodePurble:
The `writer` parameter for `bibtexparser.dump()` and `bibtexparser.dumps()` seems to be implemented, yet the docs say it isn't yet. I have a PR lined up that removes this. Please check if I have missed anything.

MiWeiss:
Fixed in @330

PR #330, opened 2022-10-09
remove `not implemented` tag from dump() & dumps()

CodePurble:
This PR is with reference to #329. Please check if there are any inaccuracies.


MiWeiss (approved):
Nice catch. Thanks

Issue #327, opened 2022-10-03
getnames edge-case


tdegeus:
```python
>>> bibtexparser.customization.getnames(["{G{\'{e}}rard} {Ben Arous}"])
["Arous}, {G{\'{e}}rard} {Ben"]
```
whereas intended:
```python
>>> bibtexparser.customization.getnames(["{G{\'{e}}rard} {Ben Arous}"])
["{Ben Arous}, {G{\'{e}}rard}"]
```

Just to note: the current output results in illegal LaTeX.

MiWeiss:
Thanks.

While most effort is on the dev of v2, I guess this one would be worth fixing for v1. @tdegeus would have the capacity to open a PR with a fix?

PR #331, opened 2022-10-09
:bug: Do not break curly bracket enclosings in `customizations.get_names`

QuantumNovice:
Now the output will be correct. The code splits on like the first level of nested {} so stuff works out.
```python
>>> bibtexparser.customization.getnames(["{G{\'{e}}rard} {Ben Arous}"])
["{Ben Arous}, {G{\'{e}}rard}"]
```

Fixes #327

PR #334, opened 2022-10-11
:bug: `customization.getnames`: respect protected names

### LINKED ISSUE #9: getnames incorrectly parsing names with braces
Sometimes names have braces, and these aren't understood by getnames. For example,

`getnames(['A. {Delgado de Molina}'])` returns `['de Molina}, A. {Delgado']`

instead, it should return `['Delgado de Molina, A.']`

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/327

I worked on https://github.com/sciunto-org/python-bibtexparser/issues/327 all while not noticing https://github.com/sciunto-org/python-bibtexparser/pull/331 .

Not surprisingly the solution is the same, with the difference that the current PR only breaks at spaces outside a sequence protected by `"{...}"` in line with the rest of the implementation of `getnames`. This reduces some of the worries discussed there I believe

Update: I was reading https://github.com/sciunto-org/python-bibtexparser/pull/331 and realised that I could do things better (safer). It does account for escaped brackets too. I think that this PR should be merged in favour of https://github.com/sciunto-org/python-bibtexparser/pull/331 . BTW, it seems to fix #9 as well that I happen to find in the tests . Note that there is a panic-switch: in case of unmatched brackets the split is entirely not applied. Instead there could be an error (or warning) in my opinion. 

For reference, I took the implementation from my own code ( https://github.com/tdegeus/texplain/blob/21d0d50aa10a6a9a3ee437e0c374d088142da709/texplain/__init__.py#L80 ) licensed MIT 

@QuantumNovice @MiWeiss


MiWeiss:
Hi @tdegeus 

Thanks a lot for this PR! I guess we're very close to merging it. 

P.S. Next time you work on any issues you identify (which would be highly appreciated), please leave a corresponding remark in the issue. I'll then assign it to you, making sure that no one else works on the issue. Sorry to @QuantumNovice - I'll reply to you in your PR.

bibtexparser/tests/test_customization.py
```diff
@@ -22,6 +22,13 @@ def test_getnames(self):
                  'Jean la Tour',
                  'Jean le Tour',
                  'Mike ben Akar',
+                 'A. {Delgado de Molina}',
+                 r'M. Vign{\'e}',
+                 'Tom {de Geus}',
+                 'Tom de {G\{eus}',
+                 'Foo B{\'a}r',
+                 r'{G{\'{e}}rard} {Ben Arous}',
+                 'Incorrect {{name}',
```
MiWeiss, line 31: Small point: Could you add some test cases containing escaped brackets (e.g. `\{ ... \}`). I see in the code that you implemented it, but it would be great to have it covered in the tests.

Otherwise, this LGTM.

MiWeiss (changes requested):

tdegeus:
Indeed, I'm very sorry  @QuantumNovice . I started working on it even after your latests commit. So I was frustrated with myself and wanted to close my PR. However, the discussion you guys had over there made me realise that I actually had something to contribute on top of https://github.com/sciunto-org/python-bibtexparser/pull/331 in terms of escaped brackets, as that is something that I had solved in the passed.

bibtexparser/tests/test_customization.py
tdegeus: Done!

MiWeiss (approved):
thanks!

PR #336, opened 2022-10-12
:children_crossing: Add __eq__ and __ne__ methods to BibDatabase

### LINKED ISSUE #335: Add the equality and non-equality method between 2 bib databases
Feature request: add the `__eq__` and `__ne__` methods for https://github.com/sciunto-org/python-bibtexparser/blob/a5eeffd92d431c7d9077fc2fc16894eca7c15861/bibtexparser/bibdatabase.py#L45-L151

If you want, I can open the PR.

mscheltienne:
Closes #335
I'm not super familiar with the way you've setup this package, let me know if this works for you; or feel free to push changes directly to this branch.

### DIFF
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -71,7 +71,7 @@ class BibDatabase(object):
         """Equality == method."""
         # uses .__dict__ attribute to avoid handling attributes which might not
         # be defined, e.g. _crossref_updated
-        return isinstance(other, BibDatabase) and self.__dict__ == other.__dict
+        return isinstance(other, BibDatabase) and self.__dict__ == other.__dict__
 
     def load_common_strings(self):
         self.strings.update(COMMON_STRINGS)
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -80,10 +80,6 @@ class BibDatabase(object):
         else:
             return False
 
-    def __ne__(self, other: Any) -> bool:
-        """Different != method."""
-        return not self.__eq__(other)
-
     def load_common_strings(self):
         self.strings.update(COMMON_STRINGS)
 
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -69,17 +69,9 @@ class BibDatabase(object):
 
     def __eq__(self, other: Any) -> bool:
         """Equality == method."""
-        if (
-            isinstance(other, BibDatabase)
-            and other.entries == self.entries
-            and other.comments == self.comments
-            and other.strings == self.strings
-            and other.preambles == self.preambles
-            and other._not_updated_by_crossref == self._not_updated_by_crossref
-        ):
-            return True
-        else:
-            return False
+        # uses .__dict__ attribute to avoid handling attributes which might not
+        # be defined, e.g. _crossref_updated
+        return isinstance(other, BibDatabase) and self.__dict__ == other.__dict
 
     def load_common_strings(self):
         self.strings.update(COMMON_STRINGS)
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from collections import OrderedDict
 import logging
+from collections import OrderedDict
+from typing import Any
 
 logger = logging.getLogger(__name__)
 
@@ -65,6 +66,23 @@ class BibDatabase(object):
 
         #: List of fields that should not be updated when resolving crossrefs
         self._not_updated_by_crossref = ['_FROM_CROSSREF']
+
+    def __eq__(self, other: Any) -> bool:
+        """Equality == method."""
+        if (
+            isinstance(other, BibDatabase)
+            and other.entries == self.entries
+            and other.comments == self.comments
+            and other.strings == self.strings
+            and other.preambles == self.preambles
+        ):
+            return True
+        else:
+            return False
+
+    def __ne__(self, other: Any) -> bool:
+        """Different != method."""
+        return not self.__eq__(other)
 
     def load_common_strings(self):
         self.strings.update(COMMON_STRINGS)
--- a/bibtexparser/bibdatabase.py
+++ b/bibtexparser/bibdatabase.py
@@ -75,6 +75,7 @@ class BibDatabase(object):
             and other.comments == self.comments
             and other.strings == self.strings
             and other.preambles == self.preambles
+            and other._not_updated_by_crossref == self._not_updated_by_crossref
         ):
             return True
         else:
--- a/bibtexparser/tests/test_bibdatabase.py
+++ b/bibtexparser/tests/test_bibdatabase.py
@@ -24,6 +24,15 @@ class TestBibDatabase(unittest.TestCase):
         bib_db.entries = self.entries
         self.assertEqual(bib_db.entries_dict, bib_db.get_entry_dict())
 
+    def test_equality(self):
+        bib_db1 = BibDatabase()
+        bib_db2 = BibDatabase()
+        assert bib_db1 == bib_db2
+        bib_db1.entries = self.entries
+        assert bib_db1 != bib_db2
+        bib_db2.entries = self.entries
+        assert bib_db1 == bib_db2
+
 
 class TestBibDataString(unittest.TestCase):
 
--- a/bibtexparser/tests/test_bibdatabase.py
+++ b/bibtexparser/tests/test_bibdatabase.py
@@ -14,6 +14,10 @@ class TestBibDatabase(unittest.TestCase):
                 'author': 'Bird, R.B. and Armstrong, R.C. and Hassager, O.'
                 }]
 
+    comments = ['a', 'b']
+    strings = ['c', 'd']
+    preambles = ['e', 'f']
+
     def test_entries_list_method(self):
         bib_db = BibDatabase()
         bib_db.entries = self.entries
@@ -28,9 +32,25 @@ class TestBibDatabase(unittest.TestCase):
         bib_db1 = BibDatabase()
         bib_db2 = BibDatabase()
         assert bib_db1 == bib_db2
+        # entries
         bib_db1.entries = self.entries
         assert bib_db1 != bib_db2
         bib_db2.entries = self.entries
+        assert bib_db1 == bib_db2
+        # comments
+        bib_db1.comments = self.comments
+        assert bib_db1 != bib_db2
+        bib_db2.comments = self.comments
+        assert bib_db1 == bib_db2
+        # strings
+        bib_db1.strings = self.strings
+        assert bib_db1 != bib_db2
+        bib_db2.strings = self.strings
+        assert bib_db1 == bib_db2
+        # preambles
+        bib_db1.preambles = self.preambles
+        assert bib_db1 != bib_db2
+        bib_db2.preambles = self.preambles
         assert bib_db1 == bib_db2
 
 
--- a/bibtexparser/tests/test_bibdatabase.py
+++ b/bibtexparser/tests/test_bibdatabase.py
@@ -1,4 +1,5 @@
 import unittest
+from collections import OrderedDict
 from bibtexparser.bibdatabase import (BibDatabase, BibDataString,
                                       BibDataStringExpression)
 
@@ -15,7 +16,7 @@ class TestBibDatabase(unittest.TestCase):
                 }]
 
     comments = ['a', 'b']
-    strings = ['c', 'd']
+    strings = OrderedDict(c="c", d="d")
     preambles = ['e', 'f']
 
     def test_entries_list_method(self):
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ nosetests.xml
 
 # Vscode
 .vscode
+
+# macOS
+.DS_Store


bibtexparser/bibdatabase.py
```diff
@@ -66,6 +67,23 @@ def __init__(self):
...
+            and other.entries == self.entries
+            and other.comments == self.comments
+            and other.strings == self.strings
+            and other.preambles == self.preambles
+        ):
+            return True
+        else:
+            return False
+
+    def __ne__(self, other: Any) -> bool:
```
MiWeiss, line 83: I do not think we need this, as it equals the default behavior: https://stackoverflow.com/questions/4352244/should-ne-be-implemented-as-the-negation-of-eq. I would propose to just use the tests to make sure the default behavior is indeed applied.

bibtexparser/bibdatabase.py
```diff
@@ -66,6 +67,23 @@ def __init__(self):
         #: List of fields that should not be updated when resolving crossrefs
         self._not_updated_by_crossref = ['_FROM_CROSSREF']
 
+    def __eq__(self, other: Any) -> bool:
+        """Equality == method."""
+        if (
```
MiWeiss, line 72: I guess this can be simplified?
```
        return (
            isinstance(other, BibDatabase)
            and other.entries == self.entries
            and other.comments == self.comments
            and other.strings == self.strings
            and other.preambles == self.preambles
        )
```

bibtexparser/bibdatabase.py
```diff
@@ -66,6 +67,23 @@ def __init__(self):
...
+            isinstance(other, BibDatabase)
+            and other.entries == self.entries
+            and other.comments == self.comments
+            and other.strings == self.strings
+            and other.preambles == self.preambles
```
MiWeiss, line 77: I know the `self._not_updated_by_crossref` is quasi-static, but as it is not, I think I would like to have it included in the equation.

bibtexparser/tests/test_bibdatabase.py
```diff
@@ -24,6 +24,15 @@ def test_entries_dict_prop(self):
...
         self.assertEqual(bib_db.entries_dict, bib_db.get_entry_dict())
 
+    def test_equality(self):
+        bib_db1 = BibDatabase()
+        bib_db2 = BibDatabase()
+        assert bib_db1 == bib_db2
+        bib_db1.entries = self.entries
+        assert bib_db1 != bib_db2
+        bib_db2.entries = self.entries
+        assert bib_db1 == bib_db2
```
MiWeiss, line 34: please also test the other attributes (other than `.entries`). Obviously, the implementation is correct - but adding a test for each field can help us avoid nasty regressions :-)

MiWeiss (changes requested):
Thanks a lot for your PR. I have added a few minor issues to be fixed.

bibtexparser/tests/test_bibdatabase.py
mscheltienne: done

bibtexparser/bibdatabase.py
mscheltienne: Done, thanks for the head's up, I'll stop including it in my classes ;)

bibtexparser/bibdatabase.py
mscheltienne: Something like `self.__dict__ == other.__dict__`? Personally, I prefer the readable and explicit version when it's only a few attributes.

bibtexparser/bibdatabase.py
mscheltienne: Done, I feel like the structure around this crossref isn't very solid. Should we also include `self._crossref_updated`?

mscheltienne:
This does not account for the order..
`.entries` is a list of dict, and there is no good reason that the order of those dict in this list differs; but if for whatever reason it does, then the equality is broken even if all the entries are equal.

```
[dict(a=1, b=2), dict(c=3, d=4)] == [dict(c=3, d=4), dict(a=1, b=2)]
>>> False
```

mscheltienne:
And since dict are unhashable and unsortable.. I don't see an efficient way to cover this.

mscheltienne:
And that's actually not an edge case.
If you take a `.bib` exported from Zotero, it's not alphabetically sorted. So if you load it and then save it; the new `.bib` file is alphabetically sorted.
Now if you load the original and the re-saved `.bib`, they will not be equal as the entries are not similarly ordered.

MiWeiss:
Since 3.6 (if I recall correctly), the order of the dict is defined: AFAIK, it is the order the keys were added.

Also, what regards `bibtexparser`, this order is important, hence the equality operator should take it into account.

bibtexparser/bibdatabase.py
MiWeiss: Sure, all attributes ;-) Thanks!

bibtexparser/bibdatabase.py
MiWeiss: self.__dict__ == other.__dict_ has the advantage that the implementation does not have to change as new attributes are added to a class.

However, this is not the point I wanted to make ;-) I'm fine with explicitly listing attributes, **but** instead of `if (condition) then return True else return False`, just use `return condition`; as shown in the example in my comment.

mscheltienne:
It's not about the dict itself, the point is about the list of dictionary. Should we consider that 2 database with the same entries are equal, regardless of the order of those entries? I would tend to say yes..

```
db = ...
db.entries = [dict(ID=entry1, ...), dict(ID=entry2, ...)]
db2 = ...
db2.entries = [dict(ID=entry2, ...), dict(ID=entry1, ...)]
db1 == db2 ???
```

MiWeiss:
@mscheltienne Thanks for your updates. I added my responses to your comments directly in the corresponding conversations.

bibtexparser/bibdatabase.py
mscheltienne: Somehow I missed the example in your comment, sure that works.

MiWeiss:
> It's not about the dict itself, the point is about the list of dictionary. Should we consider that 2 databases with the same entries are equal, regardless of the order of those entries? I would tend to say yes..
> 
> ```
> db = ...
> db.entries = [dict(ID=entry1, ...), dict(ID=entry2, ...)]
> db2 = ...
> db2.entries = [dict(ID=entry2, ...), dict(ID=entry1, ...)]
> db1 == db2 ???
> ```

ahh, now I understand. Yes, this matters as well. `db1` should not be equal to `db2` in that case.

MiWeiss:
Nice observation. My arguments why they should still be non-equal: 
- The order of entries can be relevant when dumping (see e.g. https://github.com/sciunto-org/python-bibtexparser/pull/317) and equal databases should give an equal output.
- The load-dump-load roundtrip is not `just` a roundtrip. Dumping does by default sort the entries of the db - leading to a different `.bib` file.
- Users may call entries by their list index. Equal databases should provide the same entry for the same index.

Agree?

mscheltienne:
If there is a way to dump without the sorting, then yes it makes sense; even if it's not the default dump mode.
But changing the sorting mode from the default is super hidden.. I guess it's something along the lines of:

```
from bibtex import dump
from bibtexparser.bwriter import BibTexWriter, SortingStrategy

writer = bwriter.BibTexWriter()
writer.display_order_sorting = SortingStrategy.PRESERVE

with open("bibtex.bib", "w") as bibtex_file:
    dump(bibtex_database, bibtex_file, writer)
```

Thus, a couple of propositions.
- As requested, `==` should return True only if the databases are sorted equally. 
- If they are equal but not sorted equally, let's at least tell the user with a logging message, level `WARNING`?
- Simplify dumping with a different sorting mode. How about an additional argument in `bibtexparser.dump`: `order`, equal to either of the 3 strategies implemented. I would set this argument to `str` and use a mapping between the `str` and `SortingStrategy` since it would be simpler to use.

With this last point, the code snippet above is simplified to:

```
from bibtex import dump

with open("bibtex.bib", "w") as bibtex_file:
    dump(bibtex_database, bibtex_file, order="preserve")
```

WDYT?

mscheltienne:
Regarding the last point, the idea is to make easily accessible the most-used attributes from the Writer directly in the `dump` function, while retaining advanced formatting as Writer attributes. Especially the sorting strategies, associated with the un-documented `SortingStrategy` class.

I guess the 2 most-used attributes might be:
- `order_entries_by`, to order entries.
- `display_order_sorting`, to order fields (`field` is the name I found for the key: value pairs in each entry on the BibTex documentation). 

Thus, with those 2 attributes exposed at the `dump` level, it could be:

```
def dump(
    bib_database: BibDatabase, 
    bibtex_file: TextIO, 
    writer: Optional[BibTexWriter] = None,
    order_entries: Optional[Tuple[str]] = None,
    order_fields: Optional[str] = None,
):
    pass
```

MiWeiss:
Thanks @mscheltienne for your detailed comments. I am going to answer them one-by-one:

> As requested, == should return True only if the databases are sorted equally.

Great!

> If they contain the same entries but are not sorted equally, let's at least tell the user with a logging message, level WARNING?

I'm a bit hesitant to add a quite expensive check only for logging. Even then, I would see it as info at max. Note that currently `1.x` versions' loggings is quite limited anyways - one of the many things we're addressing in v2.

> Simplify dumping with a different sorting mode. How about an additional argument in bibtexparser.dump: order, equal to either of the 3 strategies implemented. I would set this argument to str and use a mapping between the str and SortingStrategy since it would be simpler to use.

I suggest taking this discussion into a separate issue, as it is not directly related to the `__eq__` PR.  Would you mind opening a new issue - including the proposition you made above?


Please note that what regards bibtextparser v1.x, which was idle for multiple years, but depended upon by a large number of projects, I am very conservative with code changes (e.g. I merge only backward compatible stuff, and nothing that adds too much new technical dept (~bugs)). That said, `v2` is under active development, and there, we're gonna attack the big changes - check the pinned issue if you're interested in contributing.

mscheltienne:
For the logging, I agree the check is theoretically expensive, but IMO it's unlikely that we are dealing with large objects (a couple of hundred entries I guess?). And we could omit this check if the `len` is large, e.g. above 10k. WDYT?

I'll open a separate issue for the last point and x-ref with the v2 issue. Seems like a safer place to add this, up to you to decide if you want to rework the writing API in a more user-friendly manner ;) The current one looks complete, but it took me a bit of digging in the code to figure out what was doing what and how to configure the writer. It wasn't clear that I had to change the attributes after initializing the object.

I quickly wrote this package https://github.com/mscheltienne/bibclean which depends on yours to clean-up `.bib` files provided to sphinx builds and to check if the files are processed in CIs (a similar approach to `black`, `isort` but for `.bib`). I'll keep an eye on v2, in case some changes could be useful on my side.

MiWeiss:
> For the logging, I agree the check is theoretically expensive, but IMO it's unlikely that we are dealing with large objects (a couple of hundred entries I guess?). And we could omit this check if the len is large, e.g. above 10k. WDYT?

I'm still a bit reluctant, but the threshold (even though I would choose it below 10k) does certainly make it better :-) My favorite solution would still be to add some description of the `==` behavior to the class docstring and not perform any check during `__eq__` at all.  Am I missing a common use case where this could have severe consequences?

> Seems like a safer place to add this, up to you to decide if you want to rework the writing API in a more user-friendly manner

`v2` is a complete re-write and yes, better usability (type hints, classes, documentation, no config via hidden args (which are the standard in v1) , ...) is certainly a primary reason why we're doing the re-write. :smile: 

>  https://github.com/mscheltienne/bibclean

Looks nice - looking forward to seeing where the project goes. There's still room for a pyhon equivalent of the npm `bibtex-tidy` :rocket:

mscheltienne:
> I'm still a bit reluctant, but the threshold (even though I would choose it below 10k) does certainly make it better :-) My favorite solution would still be to add some description of the == behavior to the class docstring and not perform any check during __eq__ at all. Am I missing a common use case where this could have severe consequences?

If 2 databases have the same entries I would expect them to be equal. I had a use case in mind a couple of days ago when I wrote `bibclean`, but I forgot since then 😓 + I'm a bit worried about this I/O round trip.
The logging message is more a bandaid than anything, let's skip it here if you are against it; but please give some thoughts on how you want to compare 2 bib databases in v2. A couple of thoughts:

- I think 2 databases with the same entries should be equal, regardless of order.
- But order is important for selection by index -> how about skipping this issue entirely? A BibTex file is always defined with a `cite-key` and its associated value. Sounds like a dictionary.
- With a `dict`, order is now only defined at the writer level.

I might be overseeing something, but for `v2`, I don't see the need for a list of entries.

---------------

Ah! So it already existed `bibtex-tidy`! I looked for an hour on how to check/clean `.bib` files in CI workflows and couldn't find anything 😅 That one would have probably done the job just fine! At least it's in Python too now :no_mouth:

MiWeiss:
Ok :-) Thanks for all your input and comments! Much of it really highlights the need for `v2` finally leaving some of the old `1.x` debits behind.

`v2` is going to follow a quite different paradigm - I'll try to remind tagging you when I have some documentation ready ;-)

For now, what regards this PR, there are still a bunch of review comments to address, if I'm not mistaken?

mscheltienne:
Ahah no stress at all! There are indeed some remaining changes to make, I'll do it this evening from a different computer. 
Looking forward to v2 ;)

bibtexparser/bibdatabase.py
mscheltienne: Ended up using `.__dict__` as it's simpler since some attributes are not defined in `__init__` but by other methods.

bibtexparser/tests/test_bibdatabase.py
mscheltienne: I've checked a bit the doc, but honestly, I couldn't find an example. Is this correct:
```
    comments = ['a', 'b']
    strings = OrderedDict(c="c", d="d")
    preambles = ['e', 'f']
```

Especially for strings, are the values strings or maybe `BibDataString`?

mscheltienne:
@MiWeiss I left comments on the reviews.

mscheltienne:
I didn't notice the typo and the failing tests.. fixed.

bibtexparser/tests/test_bibdatabase.py
MiWeiss: Excuse my late reply. I'll need to dig into this a bit more. Unfortunately, I currently don't have much time and also must prioritize v2. I'll get back to this PR as soon as possible. Thanks for your understanding!

bibtexparser/bibdatabase.py
```diff
@@ -66,6 +67,12 @@ def __init__(self):
         #: List of fields that should not be updated when resolving crossrefs
         self._not_updated_by_crossref = ['_FROM_CROSSREF']
 
+    def __eq__(self, other: Any) -> bool:
+        """Equality == method."""
```
MiWeiss, line 71: This docstring does not really provide a useful description.

State what's considered equal (e.g. same content, same order, ...)

MiWeiss (changes requested):
Excuse the delayed review. Unfortunately, I am still unable to merge this; it has potential side effects on very many existing applications and we thus have to be extra certain when merging this...

If you are still interested in working on this:

1. Could you end a bunch of end-to-end test cases, where (identical and non-identical) BibTeX strings are parsed and checked for (non-)equality?
2. Improve the docstrings (I added a new review comment). Also add a short description to the BibDatabase docstring about the equality behavior.

MiWeiss:
Closing as stale, and I am still not 100% convinced we need / want this, thus I am not labelling this for someone to take over.

PR #351, opened 2023-01-13
Fix getnames to not tidy quoted author names

noahkw:
The implementation of getnames currently does not special-case quoted author names so that a company name like `{Google Inc}` gets added to the tidynames list as `Inc}, {Google`. This PR fixes that so that `{Google Inc}` is added to the list unchanged.


tdegeus:
The functionality is already there, see https://github.com/sciunto-org/python-bibtexparser/pull/334 . I think the issue should be sought in the `if` `else`. Could you add your not working case to https://github.com/sciunto-org/python-bibtexparser/blob/master/bibtexparser/tests/test_customization.py such that we can investigate from there?

noahkw:
Hi, it seems like currently only an extra comma and space are being added to the end of the name, so `"{Company Inc.}"` becomes `"{Company Inc.}, "`

PR #373, opened 2023-05-26
:white_check_mark: Implement test case provided in #272

### LINKED ISSUE #272: Failing in converting strings to unicode
Suppose you have this bibtex file (call it `fail.bib`):

```bibtex
@article{a,
  author = {One Two and Three{\'\i}abc-Four{\'\i}def},
}
```

The example program: 
```python
#! /usr/bin/python3
#
import bibtexparser
from bibtexparser.bparser import BibTexParser
from bibtexparser.customization import convert_to_unicode
bf=open("fail.bib")
bib_database = bibtexparser.bparser.BibTexParser(common_strings=True,
                                                 customization=convert_to_unicode
                                                ).parse_file(bf)
bf.close()
print(bib_database.entries)
```

produces the following output:

```
{'author': 'One Two and Threeı́abc-Four\\d́ef', 'ENTRYTYPE': 'article', 'ID': 'a'}]
```

which is evidently wrong. It seems that `\i` is converted to `ı` (dotless i) too early, and then `\'ı` creates havoc.

I am not sure what the solution could be, because I do not follow the code very well --- quite too complex for my skill level, I fear.

MiWeiss:
closes #272

### DIFF
--- a/tests/middleware_tests/test_latex_encoding.py
+++ b/tests/middleware_tests/test_latex_encoding.py
@@ -41,6 +41,11 @@ from tests.middleware_tests.middleware_test_util import (
             r"See https://human_resources.com",
             id="Keep special chars in url",
         ),
+        pytest.param(
+            r"One Two and Three{\'\i}abc-Four{\'\i}def",
+            "One Two and Threeíabc-Fourídef",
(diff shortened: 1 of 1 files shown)


Rmano:
Thanks!

PR #377, opened 2023-05-31
solve name conflict

bajinsheng:
The `write_string` name is used twice in entrypoint,py and writer,py

### DIFF
--- a/bibtexparser/entrypoint.py
+++ b/bibtexparser/entrypoint.py
@@ -5,7 +5,7 @@ from .library import Library
 from .middlewares.middleware import Middleware
 from .middlewares.parsestack import default_parse_stack, default_unparse_stack
 from .splitter import Splitter
-from .writer import BibtexFormat, write_string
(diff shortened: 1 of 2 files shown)


MiWeiss:
Thanks. I was aware I used the same name, but then I guess I screwed up the `__init__.py`.  Anyways, your change makes a lot of sense, provided that the `writer.py` only has a method to "write to the string" (and not to file, which is handled in `entrypoint.py`.

I am happy to merge this PR once the tests are green: Please make sure to fix the failing tests, i.e., run `black` and `isort --profile black` on the files you changed to fix the linting errors, and make sure that the tests do not call `writer.write_string` anymore.

In a later PR, we'll also have to make sure the test suite finds such errors :stuck_out_tongue_closed_eyes:  Not sure yet how that could have slipped :smile:

MiWeiss (approved):

MiWeiss:
Thanks a lot for your report and fix!

PR #380, opened 2023-06-02
fix: Exception when _treat_failed_block is called

m-stoeckel:
In `_treat_failed_block()`, `PARSING_FAILED_COMMENT` is formatted using `PARSING_FAILED_COMMENT.format({"n": lines})`. However, `str.format()` requires key-word arguments, not a dictionary. 

Thus `_treat_failed_block()` will always throw a `KeyError` exception.

This PR should fix the issue.


m-stoeckel:
I agree, tests (in general) for blocks that fail to parse would be great. I will see what I can do.

m-stoeckel:
Okay something broke on my end, I will open a new RP.

PR #387, opened 2023-07-06
:bug:  Handle `None` and Collections in BlockMiddleware.transformBlock

### LINKED ISSUE #386: BlockMiddlewareTransformer does not handle `None` or collections
According to documentation, transformers implementing `BlockMiddlewareTransformer` can return `None`, a single entry, or a collection of entries for the `transform_block` function. But returning anything other than a single entry results in a crash. The problem seems to stem from this code in `middleware.py`:

```python3
    def transform(self, library: "Library") -> "Library":
        # TODO Multiprocessing (only for large library and if allow_multi..)
        blocks = [self.transform_block(b, library) for b in library.blocks]
        return Library(blocks=blocks)
```

It doesn't handle cases where `transform_block` does not return an entry.

mlutze:
This fixes #386.

For the tests I stole the sample blocks from `test_sorting_blocks.py`.

### DIFF
--- a/bibtexparser/middlewares/middleware.py
+++ b/bibtexparser/middlewares/middleware.py
@@ -83,9 +83,16 @@ class BlockMiddleware(Middleware, abc.ABC):
             # Case 2: A single block. Add it to the list.
             elif isinstance(transformed, Block):
                 blocks.append(transformed)
-            # Case 3: Something else. Assume it's a collection and append them all.
-            else:
+            # Case 3: A collection. Append all the elements.
+            elif isinstance(transformed, Collection):
+                # check that all the items are indeed blocks
+                for item in transformed:
+                    if not isinstance(item, Block):
+                        raise TypeError(f"Non-Block type found in transformed collection: {type(item)}")
                 blocks.extend(transformed)
+            # Case 4: Something else. Error.
+            else:
+                raise TypeError(f"Illegal output type from transform_block: {type(transformed)}")
         return Library(blocks=blocks)
 
     def transform_block(
--- a/bibtexparser/middlewares/middleware.py
+++ b/bibtexparser/middlewares/middleware.py
@@ -77,10 +77,15 @@ class BlockMiddleware(Middleware, abc.ABC):
         blocks = []
         for b in library.blocks:
             transformed = self.transform_block(b, library)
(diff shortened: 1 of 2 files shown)


mlutze:
@MiWeiss Is there anything more needed before this can be merged?

MiWeiss:
Please apologize my delay, I had not noticed the PR previously.  Will try to do a review within the next couple of days.

Thanks already for opening the PR :rocket:

bibtexparser/middlewares/middleware.py
```diff
@@ -74,7 +74,18 @@ def metadata_key() -> str:
...
+        for b in library.blocks:
+            transformed = self.transform_block(b, library)
+            # Case 1: None. Skip it.
+            if transformed is None:
+                pass
+            # Case 2: A single block. Add it to the list.
+            elif isinstance(transformed, Block):
+                blocks.append(transformed)
+            # Case 3: Something else. Assume it's a collection and append them all.
+            else:
```
MiWeiss, line 87: I am considering if we should now use the chance to double-check if "something else" is really a collection of blocks, and throw an exception otherwise. I guess that would be a relatively easy and runtime-wise cheap check, but heavily simplify debugging if any (e.g. custom) middleware returns something unexpected. 

What do you think?

MiWeiss (changes requested):
Really nice, thanks a lot. I especially like the tests!

I left one comment for your consideration. Please don't hesitate to disagree if you see any reason to ;-)

bibtexparser/middlewares/middleware.py
mlutze: I think it's a good idea. I'll make the change.

MiWeiss:
Great,  thanks! 

I guess the three new test cases could be refactored into a single `@pytest.mark.parameterized` test function, thus keeping the test suite simpler to read and maintain. Or am I missing something?

Would you mind doing this refactoring? Otherwise I can take care of it. 

Afterwards,  this will be ready to be merged...

MiWeiss (approved):
Great, thanks a lot.  
There's some linting missing, after which this will be merged :-)

MiWeiss:
Thanks a lot @mlutze for the PR. Nice work!

PR #395, opened 2023-08-31
Added option to select encoding of .bib file in parse_file()

himcraft:


MiWeiss:
Thanks @himcraft for the PR. 

The code change looks good, but would you mind adding a few tests? I know that the change is rather trivial, thus the purposes of the tests would not be to verify if you change is correct (I am convinced that it is), but instead serve to prevent from regression, i.e., from unintentionally breaking this function in the future.

Thanks!

himcraft:
Added `test_encoding.py` and a `gbk_test.bib` file in `tests/`.
This is my first attempt at writing tests for a Python project. Please feel free to point out any errors or suggest improvements :)

tests/test_encoding.py
```diff
@@ -0,0 +1,17 @@
+"""Testing the parse_file function."""
+
+import pytest
+
+from bibtexparser import parse_file, writer
+
+
+def test_gbk():
+    library = parse_file("tests/gbk_test.bib", encoding="gbk")
+    lines = writer.write(library).splitlines()
```
MiWeiss, line 10: I would suggest not making the assertion on the "re-written" bibtex, but on the parsed entities instead.

e.g. `assert library.entries[0]["author"] == ...`

That way, we can keep the test independent of the writer.

tests/test_encoding.py
```diff
@@ -0,0 +1,17 @@
+"""Testing the parse_file function."""
```
MiWeiss, line 1: can we re-name this file to `test_entrypoint` and the module docstring accordingly - there's likely to be added more tests to it eventually.

tests/gbk_test.bib
```diff
@@ -0,0 +1,6 @@
+@ARTICLE{����2013,
```
MiWeiss, line 1: Please move this file into a (new) `tests/resources` folder.

MiWeiss (changes requested):
Thanks a lot for your commit.

Your test would certainly already be sufficient as a regression test. I still have left some minor comments, mostly to keep the test atomic and the test organisation (read file structure) consistent with the remainder of the test suite.

Afterwards, this will be ready to be merged :rocket: Thanks!

CONTRIBUTING.md
```diff
@@ -28,5 +28,5 @@ Issues are labelled `v1` and `v2`, correspondingly.
 To install the dev dependencies, run `pip install -e .[test,lint,docs]` from within the cloned repository. Then:
 
 - To test your code, run `pytest .`
-- To lint your code (required for CI/CD to pass), run: `black bibtexparser tests docs && isort bibtexparser tests docs --profile black
+- To lint your code (required for CI/CD to pass), run: `black bibtexparser tests docs && isort bibtexparser tests docs --profile black`
```
MiWeiss, line 31: Nice catch ;-)

MiWeiss (approved):
lgtm!

Thanks a lot for your contribution. If there's any more problems or missing features you spot, please do not hesitate to report them.

PR #405, opened 2023-09-21
Add encoding to file writing function

### LINKED ISSUE #403: Writing doesn't work for umlauts (likely UTF-8 formatting problem)
**Describe the bug**
When I create a library with unicode characters (tested with umlauts), the export will not show them properly.

**Reproducing**

Version: Most current from PyPI

Code:
```python
import bibtexparser
from bibtexparser import *

bib_library = bibtexparser.Library()

fields = []

fields.append(bibtexparser.model.Field("author", "ö"))

entry = bibtexparser.model.Entry("ARTICLE", "test", fields)

bib_library.add(entry)

print(bib_library.entries_dict)

bibtexparser.write_file("my_new_file.bib", bib_library)
```

Bibtex:
```
@ARTICLE{test,
	author = {�}
}

```

**Workaround**
No; possibly generating a string first and then writing that manually will work (similarly to the workaround for #394).

Please tick all that apply:

- [ ] I would be willing to to contribute a PR to fix this issue.
- [ ] This issue is a blocker, I'd be greatful for an early fix.

claell:
Fixes #403

### DIFF
--- a/bibtexparser/entrypoint.py
+++ b/bibtexparser/entrypoint.py
@@ -125,6 +125,7 @@ def write_file(
     parse_stack: Optional[Iterable[Middleware]] = None,
     append_middleware: Optional[Iterable[Middleware]] = None,
     bibtex_format: Optional[BibtexFormat] = None,
+    encoding: str = "UTF-8",
 ) -> None:
     """Write a BibTeX database to a file.
 
@@ -134,7 +135,8 @@ def write_file(
                         If None, a default stack will be used.
     :param append_middleware: List of middleware to append to the default stack.
                         Only applicable if `parse_stack` is None.
-    :param bibtex_format: Customized BibTeX format to use (optional)."""
+    :param bibtex_format: Customized BibTeX format to use (optional).
+    :param encoding: Encoding for the .bib file. Default encoding is "UTF-8"."""
     bibtex_str = write_string(
         library=library,
         unparse_stack=parse_stack,
@@ -142,7 +144,7 @@ def write_file(
         bibtex_format=bibtex_format,
     )
     if isinstance(file, str):
-        with open(file, "w") as f:
+        with open(file, "w", encoding=encoding) as f:
             f.write(bibtex_str)
     else:
         file.write(bibtex_str)


claell:
Right now, this is just adapted from #395. Not tested, just quickly done in GitHub editor.

MiWeiss (changes requested):
Thanks for the PR. Same as in #395, we would need some tests, primarily to prevent regressions. After those are added, the PR can be merged.

claell:
Sorry, am rather busy right now, so not sure when (or if) I'll have time for that.

MiWeiss:
> Sorry, am rather busy right now, so not sure when (or if) I'll have time for that.

Thanks for the comment. I'll mark the PR to be ready for anyone to continue working on this. 

If anyone is willing to do so, or if @claell resumes his work: Please comment here to avoid having two people working on the same thing at the same time.

p.s.  Continuing to work on the PR can be done as follows: Add the fork of @claell as additional `git remote`, pull the branch, and then push it to your own fork and open a new PR, stating that that one is supposed to replace this one.

agriyakhetarpal:
Hi, we're looking to add `python-bibtexparser` as a dependency for the citations workflow of our Python package soon – I thought it would be nice to contribute as well, perhaps by unblocking a few PRs; such as this one and suchlike. 

I am happy to take over from @claell and write a few tests, though the issue is that I fail to reproduce the original issue (#403) at this time:

Using the core functionality, i.e., without any customizations to the formatting options for writing, I am able to write an entry:

```python
import bibtexparser
from bibtexparser import *
bib_library = bibtexparser.Library()
fields = []
fields.append(bibtexparser.model.Field("author", "ö"))
entry = bibtexparser.model.Entry("ARTICLE", "test", fields)
bib_library.add(entry)
bibtexparser.write_file("my_new_file.bib", bib_library)
```

i.e., the same MWE as previously reported in https://github.com/sciunto-org/python-bibtexparser/issues/403#issue-1905224824, and parsing `my_new_file.bib` seems to work without issues.

> ### my_new_file.bib
```bib
@ARTICLE{test,
	author = {ö}
}
```
and reading this programmatically:
```python
import bibtexparser
library = bibtexparser.parse_file("my_new_file.bib")
```
Therefore, `library.entries_dict` returns
```python
{'test': Entry(entry_type=`article`, key=`test`, fields=`[Field(key=`author`, value=`ö`, start_line=1)]`, start_line=0)}
```
which I am able to write to a new `.bib` file via the `bibtexparser.write_file()` method without any loss of data. Has this been fixed, or cannot be reproduced, or there is a different method of looking at the error that I may have missed in oversight?

MiWeiss:
Hi @agriyakhetarpal 

Great, thanks a lot for taking over! 

Strange to hear that the problem cannot be reproduced, I am not aware of any changes we merged recently (although I have not checked ;-) ). As we're talking about encoding, I would not be surprised if the behavior was somewhat system-dependent, which could explain the issue at hand.

In either case, I think we should still merge this PR - to keep the interface consistent and generally applicable. While it's not ideal to do so without reproducing the problem above, I'd suggest implementing tests similar to #395 - this should, at the very least, protect us from some regressions. Do you agree?

> I thought it would be nice to contribute as well

That would be amazing. I'll be happy to help you along the way wherever I can (feedback, reviews, ...)

agriyakhetarpal:
> As we're talking about encoding, I would not be surprised if the behavior was somewhat system-dependent, which could explain the issue at hand.
>
> In either case, I think we should still merge this PR - to keep the interface consistent and generally applicable. While it's not ideal to do so without reproducing the problem above, I'd suggest implementing tests similar to #395 - this should, at the very least, protect us from some regressions. Do you agree?

> I'll be happy to help you along the way wherever I can (feedback, reviews, ...)

I could not agree more. I am aware systems like Windows choose CP1252 by default for writing to files if UTF-8 isn't specified – maybe the reason I'm not seeing this error is because I'm on macOS? 

Is it fine if we can discuss things here itself before I proceed to write a PR? I couldn't wrap my head around the source code for the middlewares and customisers for now—very custom engineering TBF—but I did manage to write a simple test for `test_writer.py`, as follows.

```python
def test_write_article_with_umlauts():
    entry_block = Entry(
        entry_type="article",
        key="myKey",
        fields=[
            Field(key="title", value='"myTitle"'),
            Field(key="author", value='"Müller, Max"'),
        ],
    )
    library = Library(blocks=[entry_block])
    string = writer.write(library)
    print(library)
    print(library.entries_dict)
    print(string)
    assert string == '@article{myKey,\n\ttitle = "myTitle",\n\tauthor = "Müller, Max"\n}\n'
```

I would appreciate feedback on this, afterwards we should be able to `pytest`-parameterize this with a few other characters (ö, ä, ë, and other diacritics you can think of!)

MiWeiss:
Hi @agriyakhetarpal 

Your suggested test, as I read it, would not actually test the method targeted in this PR.

Instead, you would have to test the method `bibtexparser.write_file` and included the newly added encoding parameter. This will write a new bibtex file. You could then check if this file contains the expected content in the expected encoding.

agriyakhetarpal:
>Instead, you would have to test the method bibtexparser.write_file and included the newly added encoding parameter. This will write a new bibtex file. You could then check if this file contains the expected content in the expected encoding.

Makes sense. I improved the test such that it parses a given BibTeX string with umlauts symbols, writes to a temporary file, and reads from it; as follows:

```python
def test_write_file_with_umlauts():
    bibtex_str = """@article{umlauts,
    author = {Müller, Hans},
    title = {A title},
    year = {2014},
    journal = {A Journal}
    }"""
    library = parse_string(bibtex_str)
    with tempfile.NamedTemporaryFile(mode="w", encoding="utf-8") as f:
        write_file(f, library, encoding="utf-8")
        f.seek(0)
        library = parse_file(f.name, encoding="utf-8")
    assert library.entries[0]["author"] == "Müller, Hans"
    assert library.entries[0]["title"] == "A title"
    assert library.entries[0]["year"] == "2014"
    assert library.entries[0]["journal"] == "A Journal"
```

I opted to use temporary files so as to not create any clutter. Does this look fair enough? I can then write a few more tests or parameterize this as needed, perhaps with a few more characters.

auge:
the problem still exists on Windows (as outlined above).
as a work-around one can also set an environment variable (using your favorite shell or in the system...):
`export PYTHONUTF8=1`

amcar-dtu:
hi, I have also experienced a similar issue, namely:

```
  File "C:\Users\usr\documents\latex\doc\scripts\bib_parser.py", line 44, in <module>
    clean_bib()
    ~~~~~~~~~^^
  File "C:\Users\usr\documents\latex\doc\scripts\bib_parser.py", line 39, in clean_bib
    bibtexparser.write_file(bibtex_file, bibtex_database)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\usr\AppData\Local\miniconda3\Lib\site-packages\bibtexparser\entrypoint.py", line 160, in write_file
    f.write(bibtex_str)
    ~~~~~~~^^^^^^^^^^^^
  File "c:\Users\usr\AppData\Local\miniconda3\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'charmap' codec can't encode character '\u2212' in position 2333: character maps to <undefined>
```

The fix that worked was actually the one implemented in #405 .

The workaround from @auge also works good (especially if one sets it in the `launch.json` configuration as a `"env"` key)

Issue #411, opened 2023-10-03
handling of new blocks that aren't on a new line


benlogan:
**Describe the bug**

V1 handled input files where the new blocks (indicated by @) aren't necessarily on a new line.
V2 attempts to ignore these, but instead get's into a real mess, because they end up being 'nested' within other entries. If they were properly ignored (i.e. parsed and dropped), then it would kind of be ok and consistent, but they make it into the raw data for each entry.
This leads to all sorts of problems. They will reappear, for example, when you write the library back out to a file later! 
Very confusing.
IMHO, they should be treated as valid blocks and handled properly. I've tested a fix for this locally and it appears to work OK.

**Reproducing**

Version: latest

Code:

parse any file with inline new blocks, e.g.;
parsed_lib = bibtexparser.parse_file(filename)

Bibtex:
```
@article{KUEH2023S98,
title = {Myocardial Characterisation Using Delayed Dual-Energy Cardiac Computed Tomography},
journal = {Heart, Lung and Circulation},
volume = {32},
pages = {S98},
year = {2023},
note = {Abstracts for the Cardiac Society of Australia and New Zealand Annual Scientific Meeting (New Zealand) 2023, 15 - 17 June 2023, Auckland, New Zealand},
issn = {1443-9506},
doi = {https://doi.org/10.1016/j.hlc.2023.04.262},
url = {https://www.sciencedirect.com/science/article/pii/S1443950623004390},
author = {S.-H. Kueh and J. Benatar and R. Stewart}
}@INPROCEEDINGS{9837531,
  author={Hassan, Mona Bakri and Saeed, Rashid A. and Khalifa, Othman and Ali, Elmustafa Sayed and Mokhtar, Rania A. and Hashim, Aisha A.},
  booktitle={2022 IEEE 2nd International Maghreb Meeting of the Conference on Sciences and Techniques of Automatic Control and Computer Engineering (MI-STA)}, 
  title={Green Machine Learning for Green Cloud Energy Efficiency}, 
  year={2022},
  volume={},
  number={},
  pages={288-294},
  doi={10.1109/MI-STA54861.2022.9837531}}@ARTICLE{9372936,
  author={Hu, Ning and Tian, Zhihong and Du, Xiaojiang and Guizani, Nadra and Zhu, Zhihan},
  journal={IEEE Transactions on Green Communications and Networking}, 
[...]

PR #412, opened 2023-10-03
Update splitter.py

benlogan:
support for new entries on the same line

has been tested locally with files that were previously triggering the bug mentioned in 
https://github.com/sciunto-org/python-bibtexparser/issues/411
but hasn't been unit tested

I'm still just getting my head around how this library works - so it's worth checking carefully.

but I think this is an important change - I spent hours trying to work out what was causing corruption in my exported files

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -52,7 +52,7 @@ class Splitter:
         # By default, we assume that an implicit comment is started
         #   at the beginning of the file and after each @{...} block.
         #   We then ignore empty implicit comments.
-        self._implicit_comment_start_line = self._current_line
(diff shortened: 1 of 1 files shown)


MiWeiss:
Thanks for the PR.

Besides the fact that there are still failing tests, this PR won't be mergeable in its current form: The "every block starts on a new line" is encoded in a variety of places, and extra care as well as significant testing will have to be added when changing this behavior.

I have myself played around a bit with changing this, but am unfortunately quite limited in time at the moment. Thus, opening my own PR for this will not happen very soon. Thus, @benlogan, if you intend to continue working on this PR, go ahead. Just drop a message here to make sure we're not both working on this at the same time.

MiWeiss:
(closed by accident - immediately reopened)

MiWeiss:
Replaced by #416

Issue #372, opened 2023-05-26
Support biber comments


MiWeiss:
Extracted from #133 

```
@ARTICLE{bla,
    author={Fartsy},
    %title={Title},
}
```

I see two options to handle this:

1. A middleware to filter out fields starting with %
2. Adapting the splitter with a custom flag to ignore lines starting with `%` or whitespace followed by `%`

In either case, for within-entry comments, I don't think we should create a Comment block in the library.

Option one is more in line with v2 architecture, and it prevents from flags on the splitter, but inputs like this will fail to parse if %-comments are only handled in a  middleware.

```
@ARTICLE{bla,
    %somecomment
    author={Fartsy},
    title={Title},
}
```

claell:
A standalone, validated implementation is available in draft PR #585: https://github.com/sciunto-org/python-bibtexparser/pull/585. It parses line-leading percent comments structurally, keeps them out of active fields, and preserves them explicitly through canonical writes and field edits.

PR #416, opened 2023-11-04
Fix blocks not starting on new lines (#411)

zepinglee:
This PR modifies the `@` marker regex so that it doesn't necessarily appear at a new line. It also passes all the tests.

### DIFF
--- a/tests/splitter_tests/test_splitter_basic.py
+++ b/tests/splitter_tests/test_splitter_basic.py
@@ -319,3 +319,59 @@ def test_handles_duplicate_strings():
     assert isinstance(
         lib.failed_blocks[0].ignore_error_block, bibtexparser.model.String
     )
+
+
+blocks_not_starting_on_new_lines = """
+@article{KUEH2023S98,
+title = {Myocardial Characterisation Using Delayed Dual-Energy Cardiac Computed Tomography},
+journal = {Heart, Lung and Circulation},
+volume = {32},
+pages = {S98},
+year = {2023},
+note = {Abstracts for the Cardiac Society of Australia and New Zealand Annual Scientific Meeting (New Zealand) 2023, 15 - 17 June 2023, Auckland, New Zealand},
+issn = {1443-9506},
+doi = {https://doi.org/10.1016/j.hlc.2023.04.262},
+url = {https://www.sciencedirect.com/science/article/pii/S1443950623004390},
+author = {S.-H. Kueh and J. Benatar and R. Stewart}
+}@INPROCEEDINGS{9837531,
+  author={Hassan, Mona Bakri and Saeed, Rashid A. and Khalifa, Othman and Ali, Elmustafa Sayed and Mokhtar, Rania A. and Hashim, Aisha A.},
+  booktitle={2022 IEEE 2nd International Maghreb Meeting of the Conference on Sciences and Techniques of Automatic Control and Computer Engineering (MI-STA)},
+  title={Green Machine Learning for Green Cloud Energy Efficiency},
+  year={2022},
+  volume={},
+  number={},
+  pages={288-294},
+  doi={10.1109/MI-STA54861.2022.9837531}}@ARTICLE{9372936,
+  author={Hu, Ning and Tian, Zhihong and Du, Xiaojiang and Guizani, Nadra and Zhu, Zhihan},
+  journal={IEEE Transactions on Green Communications and Networking},
+  title={Deep-Green: A Dispersed Energy-Efficiency Computing Paradigm for Green Industrial IoT},
+  year={2021},
+  volume={5},
+  number={2},
+  pages={750-764},
+  doi={10.1109/TGCN.2021.3064683}}@ARTICLE{5445167,
+  author={Kumar, Karthik and Lu, Yung-Hsiang},
+  journal={Computer},
+  title={Cloud Computing for Mobile Users: Can Offloading Computation Save Energy?},
+  year={2010},
+  volume={43},
+  number={4},
+  pages={51-56},
+  doi={10.1109/MC.2010.98}}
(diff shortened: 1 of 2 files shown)


tests/splitter_tests/test_splitter_basic.py
```diff
@@ -319,3 +319,59 @@ def test_handles_duplicate_strings():
...
+    """Test the new blocks that are not on new lines.
+
+    Discussed at https://github.com/sciunto-org/python-bibtexparser/issues/411.
+    """
+    import bibtexparser
+
+    lib = bibtexparser.parse_string(blocks_not_starting_on_new_lines)
+    assert len(lib.blocks) == 4
+    assert len(lib.entries) == 4
+    assert lib.entries[0].entry_type == "article"
```
MiWeiss, line 374: Could you add a check asserting that the `entry.start_line` values are correct?

MiWeiss (changes requested):
Hi. Thanks a lot for opening the PR. 

I left one minor comment regarding the test above.

A more significant problem is on [line 175](https://github.com/sciunto-org/python-bibtexparser/pull/416/files#diff-9928b8e2ef0ad92b3a5aa9b829f3fb50032950bc14afd79a085b47df6df2005dL175). Currently, we discover if a new block starts with a previous one still open, which typically highlights an error in the parsed file. In such a case, we abort processing the still open block, in deliberate deviation from "strict bibtex parsing". As we only considered `@....` on the start of a new line, the risk of flagging a "false positive" was negligible, but at the same time the user experience increased drastically.

With your change, we would abort the parsing of a block when facing an unexpected `@....` anywhere, e.g, the following could not be parsed:

```
@entry{key, title="all blocks must start with @entry{"} 
```

To reduce this risk, I would love to keep the previous behavior and only abort the parsing of a block if a new block starts on a new line.  Does that make sense?

*Note:*
Another problem, but one which does not have to be fixed is that biber-style comment won't be working with this PR anymore, e.g. the following will likely result in an error.,

```
% @entry{key, 
% title="all blocks must start with @entry{"} 
```
However, we technically still don't support biber comments and a clean solution is required there anyways, thus I'd leave this to #372

tests/splitter_tests/test_splitter_basic.py
zepinglee: Sure.

zepinglee:
> A more significant problem is on [line 175](https://github.com/sciunto-org/python-bibtexparser/pull/416/files#diff-9928b8e2ef0ad92b3a5aa9b829f3fb50032950bc14afd79a085b47df6df2005dL175). Currently, we discover if a new block starts with a previous one still open, which typically highlights an error in the parsed file. In such a case, we abort processing the still open block, in deliberate deviation from "strict bibtex parsing". As we only considered `@....` on the start of a new line, the risk of flagging a "false positive" was negligible, but at the same time the user experience increased drastically.

I don't think the `_move_to_comma_or_closing_curly_bracket()` method where line 175 is located matches the original BibTeX's behavior. BibTeX only accepts strictly brace-balanced string in a literal field token (in terms of the `bibtex.web`) even when the outer delimiters are quotes. The braces or quotation marks in a field token are not treated as escaped characters. In addition, an `@` character doesn't break the field.

https://github.com/sciunto-org/python-bibtexparser/blob/770eb001349dcad48149bc101f07cfc88282d8b3/bibtexparser/splitter.py#L131-L189

In the example of `@entry{key, title="all blocks must start with @entry{"}`, the second quotation mark is in brace level 1 and it doesn't terminate the field token. The final right brace make the string balanced and BibTeX look for another quotation mark to end the field. This is verified by running BibTeX and the raised error `Illegal end of database file`. Yet `@entry{key, title="all blocks must start with @entry{"}"}` works fine with BibTeX and the `title` field is `all blocks must start with @entry{"}`.

I'm planning to rewrite this method and make another PR.

tests/splitter_tests/test_splitter_basic.py
zepinglee: I've added these assertions in d9a3245. Also note that I removed fields other than `author`, `title`, and `doi` to make less lines.

MiWeiss:
Thanks for your PR update and for your long comment.

#### First things first:
Your argument convinced me that a plain `@entry{` would be very rare, and I thus see no reason not to merge this PR anymore. We may change on this feature in the future, based on received feedback, but for now I guess merging this PR is a clear improvement of things!

#### Regarding the larger argument you made:
 This library is deliberately not 100% consistent with bibtex, but it attempts to be somewhat more fault-tolerant: We aim to be able to handle cases when there are small mistakes in the bibtex file gently, without messing up the parsing of subsequent blocks, or even prevent parsing at all. This is a request we received numerous times reagarding bibtexparser v1. Take the following example:

```
@string{myKey = {myValue))

@string{anotherString = {anotherValue}}
```

While the intention of the writer here is pretty clear, bibtex would not be happy: Note the `))` on the first line which should be `}}`. Technically, the braces are not balanced after the first line, thus not allowing for a new string on the third line. Previously, in `bibtexparser v2`, we thus aborted the parsing of a block if a line started with a "block start", e.g. a `@entry{`. This allowed a "fresh" start with that block. After merging this PR, we now do this even if the "block start" appears anywhere, not just on a new line. 

Thus, I am very happy to review any improvement you make to the parsing (and I am sure there are plenty to be made ;-)), but please keep in mind that we take a deliberate trade-off between sticking to the pure bibtex rules, and user friendly parsing of imperfect files. If in doubt, before investing too much time, feel free to ask for mine (and anyones) comment in a seperate issue before investing too much time.

#### Most importantly
 Thanks again for your contribution, and I am looking forward to your next PRs.

MiWeiss (approved):

PR #419, opened 2023-11-14
fix installation documentation

mlutze:
There seems to be a typo in the documentation for installing via pip.


MiWeiss:
Ouch! Thanks for spotting this!

Please note that the docs won't be updated until #419 is fixed. Not sure when I get to do that, hopefully soon!

MiWeiss (approved):

PR #422, opened 2023-11-23
Add merging of names in last-name-first style

mlutze:
I added a flag to `MergeNameParts` so the user can decide whether to use first-name-first or last-name-first.
I made last-name-first the default.

I slightly modified one existing test group to support the flag.

I added a new test group `test_merge_last_name_first_inverse` which ensures that merging with commas is the inverse of splitting. I skip this test for cases including a `/`, since this property doesn't hold for some of those cases.

I also had to do a bit of strange stuff with the typing imports. The flag I added is a union of literal strings, but literal string types aren't available in Python 3.7. So I try importing a literal type and if that doesn't work then I just say the type of the flag is "str", which is too general but the closest I can get

### DIFF
--- a/bibtexparser/middlewares/names.py
+++ b/bibtexparser/middlewares/names.py
@@ -5,7 +5,7 @@ Much of the code is taken from Blair Bonnetts never merged v0 pull request
 """
 import abc
 import dataclasses
-from typing import List, Tuple
+from typing import List, Literal, Tuple
 
 from bibtexparser.model import Block, Entry, Field, MiddlewareErrorBlock
 
@@ -171,15 +171,19 @@ class MergeNameParts(_NameTransformerMiddleware):
     """Middleware to merge a persons name parts (first, von, last, jr) into a single string.
 
     Name fields (e.g. author, editor, translator) are expected to be lists of NameParts.
+
+    The merging style indicates whether:
+    - the merging is done without commas in first-name-first order ("first"), or
+    - the merging is done with commas in last-name-first order.
     """
 
     def __init__(
         self,
-        last_name_first: bool = True,
+        style: Literal["last"] | Literal["first"] = "last",
         allow_inplace_modification: bool = True,
         name_fields: Tuple[str] = ("author", "editor", "translator"),
     ):
-        self._last_name_first = last_name_first
+        self.style = style
         super().__init__(allow_inplace_modification, name_fields)
 
     # docstr-coverage: inherited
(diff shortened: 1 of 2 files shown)


mlutze:
> Thanks for your work on this. Just to avoid a misunderstanding: I wont review this PR while it is marked as draft.

Understood. There are still some changes to testing and documentation I'd like to make before submitting for review.

mlutze:
@MiWeiss This is ready for review.

bibtexparser/middlewares/names.py
```diff
@@ -7,6 +7,14 @@
 import dataclasses
 from typing import List, Tuple
 
+# Literal is only available in Python 3.8+
+try:
```
MiWeiss, line 11: In the meantime, python 3.7 has reached EOL. 

I have just updated the CI on master to not use 3.7 anymore, thus feel free to remove your fallback.

tests/middleware_tests/test_names.py
```diff
@@ -860,6 +860,27 @@ def test_split_name_into_parts(name, expected_as_dict, strict):
...
+@pytest.mark.parametrize("strict", [True, False], ids=["strict", "non-strict"])
+def test_merge_last_name_first_inverse(name, expected_as_dict, strict):
+    """Tests that merging name parts using the last-name-first method
+    maintains the "semantics" of the name.
+
+    This property does not hold for certain values that contain '\\'.
+    """
+
+    # skip cases with \
+    if any("\\" in "".join(name) for name in expected_as_dict.values()):
```
MiWeiss, line 875: Can we somehow document in which cases we cannot handle this? I have struggle understanding the severity of this case.

MiWeiss (changes requested):
Hi @mlutze 

Thanks a lot for opening th e PR. This is certainly a very important contribution.

I see (and acknowledge) that the whole `\\` handling is a bit of a struggle. Still, I'd like to understand it a bit better to know whats going on.  I hope you don't mind my corresponding review ;-)

tests/middleware_tests/test_names.py
mlutze: As far as I can tell, the cases that cannot be handled are those where either the last name or "von" part ends with an odd number of `\`, since in those cases the `,` is escaped when the name parts are put back together. I add an extra `\` in those cases, which I _think_ preserves the semantics of the name, though I'm not 100% sure.

The filter here is more aggressive than what I put above though, to keep it simple. But I guess it's better if the filter captures exactly the cases for which the "inverse" property holds?

Where would you like the documentation for the edge cases?

MiWeiss: Ok, I guess that's enough of an edge case and ok to have it in the code, since I really do not think that single trailing `\`  are common enough to block this PR :-)

MiWeiss:
```suggestion
    # cases where either the last name or "von" part ends with an odd number of `\` cannot be handled, 
    # since in those cases the `,` is escaped when the name parts are put back together
    if any("\\" in "".join(name) for name in expected_as_dict.values()):
```

mlutze: Thanks. I've added the comment and made the filter a bit more lenient so only cases with an odd number of trailing `\` are discarded.

MiWeiss:
Thanks again @mlutze  for your PR, and sorry that the review took so long!

MiWeiss (approved):

PR #423, opened 2023-11-23
:bug: Fix bug in `von`-part name parsing

mlutze:
The current system has a name parsing issue around capitalization in comma-style names. This means that

```
aa bb, CC
```

gets parsed as

```
last: aa bb
first: CC
von: <empty>
```

but BibTeX parses this as

```
last: bb
first: CC
von: aa
```

I have modified the parsing code and added relevant test cases drawn from the Tame The BeaST BibTeX reference document.

Style-wise, I found the list manipulation difficult to follow; I added a helper function `rindex`, which makes it easier to understand IMO. This should probably go in a utility file or something, but I didn't see one in the project.

I believe these changes will fix at least some of the failing test cases in the name merging PR here: #422

### DIFF
--- a/bibtexparser/middlewares/names.py
+++ b/bibtexparser/middlewares/names.py
@@ -424,11 +424,17 @@ def parse_single_name_into_parts(name, strict=True):
         else:
             lcases = cases[0]
 
-            #
+            def rindex(l, x, default):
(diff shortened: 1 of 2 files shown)


MiWeiss (approved):
Looks great, thanks!

PR #427, opened 2023-12-15
Use raw strings to fix invalid escape (V1)

kno10:
The replacement string contains `\g`, which is an invalid escape in Python (and not supposed to be a python escape).


MiWeiss (approved):
Thanks :-)

Issue #439, opened 2024-01-18
Separator first names


tdegeus:
**Is your feature request related to a problem? Please describe.**
I would like to customize the symbol with which first names are joined.

**Describe the solution you'd like**
```python
bibtex_str = """
@ARTICLE{Cesar2013,
  author = {Jean César and T. W. J. de Geus},
  title = {An amazing title},
  year = {2013},
  volume = {12},
  pages = {12--23},
  journal = {Nice Journal}
}
"""

import bibtexparser
import bibtexparser.middlewares as m

layers = [
    m.SeparateCoAuthors(True),
    m.SplitNameParts(True)
]
library = bibtexparser.parse_string(bibtex_str, append_middleware=layers)

layers = [
    m.MergeNameParts(sep_first="")
    m.MergeCoAuthors(True)
]
print(bibtexparser.write_string(library, prepend_middleware=layers))
```
should then print 
```bib
@ARTICLE{Cesar2013,
  author = {J. César and de T.W.J. Geus},
  title = {An amazing title},
  year = {2013},
  volume = {12},
  pages = {12--23},
  journal = {Nice Journal}
}
```

**Use cases**
This is to dirty hack very short display while keeping the information. 

**Describe alternatives you've considered**
-

**Remaining Questions (Optional)**
Please tick all that apply:

- [x] I would be willing to to contribute a PR to fix this issue.
- [ ] This issue is a blocker, I'd be greatful for an early fix.

MiWeiss:
Hi @tdegeus 

Thanks a lot for opening this series of issues. I truly appreciate it.

This issue (as much as the other ones you opened) makes sense and I'd be happy to merge a corresponding change.

> I would be willing to to contribute a PR to fix this issue.

That would be awesome. In case #422 does not completely resolve your issue in #438, feel free to add further changes in that direction into that PR as well :-)

tdegeus:
Thanks. I will have a look soon how I can try to contribute

tdegeus:
On hold to clarify relevance and API, see https://github.com/sciunto-org/python-bibtexparser/pull/457 . Please reopen or comment at will

PR #457, opened 2024-02-06
Allow different separator first names

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/439

Should one be able to customize the other separators as well? Should the API be made immediately, rebust to this.

**Help wanted** I have no idea why the CI fails.

### DIFF
--- a/bibtexparser/middlewares/names.py
+++ b/bibtexparser/middlewares/names.py
@@ -98,14 +98,16 @@ class NameParts:
     last: List[str] = dataclasses.field(default_factory=list)
     jr: List[str] = dataclasses.field(default_factory=list)
 
-    @property
-    def merge_first_name_first(self) -> str:
-        """Merging the name parts into a single string, first-name-first (no comma) format."""
+    def merge_first_name_first(self, sep_first: str) -> str:
+        """Merging the name parts into a single string, first-name-first (no comma) format.
+
+        :param sep_first: Separator between first name parts.
+        """
         return " ".join(
             [
                 part
                 for part in (
-                    " ".join(self.first) if self.first else None,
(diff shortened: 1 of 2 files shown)


bibtexparser/middlewares/names.py
```diff
@@ -175,15 +178,22 @@ class MergeNameParts(_NameTransformerMiddleware):
...
+    :param style: Merge style (see above).
+    :sep_first: Separator between first name parts.
+    :param allow_inplace_modification: See corresponding property parent class.
+    :param name_fields: The fields that contain names.
     """
 
     def __init__(
         self,
         style: Literal["last", "first"] = "last",
+        sep_first: str = " ",
```
MiWeiss, line 191: Have you considered making this a setter to keep the constructor less verbose? I currently imagine (but please tell me what I miss) that barely anyone would use this, thus adding four more constructor params (one for each separator) seems a bit heavy. 

What are your thoughts on that?

MiWeiss (changes requested):
Hey. Could we add some test for this? 

Certainly, the feature is simple enough that your PR is trustworthy without tests, but I'd like to protect this against regressions, i.e., against future changes breaking this again by accident.

> Should one be able to customize the other separators as well? Should the API immediately be made robust to this?

I struggle seeing the usecase for this. What am I missing? If it is a frequent enough usecase, sure enough, let's add it for all separators.

tdegeus:
Your remarks make me think that, to be honest, it has been a long while since I used this dirty hack. I too doubt a bit if we should want this (as I also don't like the API that I proposed). Maybe we can put a pin it, and circle back if needed

Issue #454, opened 2024-02-02
Make behaviour entry closer to `dict`: check if field is present


tdegeus:
**Is your feature request related to a problem? Please describe.**
I would like to do something *if* a field is present in an entry. 
For example:
```python
import bibtexparser
from bibtexparser.middlewares.middleware import BlockMiddleware

bibtex_str = """
@ARTICLE{Cesar2013,
author = {Jean César and T. W. J. de Geus},
title = {An amazing title},
year = {2013},
pages = {12--23},
volume = {12},
journal = {Nice Journal}
}
"""

class Foo(BlockMiddleware):
    def transform_entry(self, entry, *args, **kwargs):
        if "journal" in entry:
            pass
        return entry

library = bibtexparser.parse_string(bibtex_str, append_middleware=[Foo(True)])
```

This now raises
```
Traceback (most recent call last):
  File "/Users/tdegeus/data/prog/src/GooseBib/t.py", line 21, in <module>
    library = bibtexparser.parse_string(bibtex_str, append_middleware=[Foo(True)])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tdegeus/miniforge3/envs/test/lib/python3.12/site-packages/bibtexparser/entrypoint.py", line 93, in parse_string
    library = middleware.transform(library=library)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tdegeus/miniforge3/envs/test/lib/python3.12/site-packages/bibtexparser/middlewares/middleware.py", line 77, in transform
    transformed = self.transform_block(b, library)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tdegeus/miniforge3/envs/test/lib/python3.12/site-packages/bibtexparser/middlewares/middleware.py", line 122, in transform_block
    return self.transform_entry(block, library)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tdegeus/data/prog/src/GooseBib/t.py", line 17, in transform_entry
    if "journal" in entry:
       ^^^^^^^^^^^^^^^^^^
  File "/Users/tdegeus/miniforge3/envs/test/lib/python3.12/site-packages/bibtexparser/model.py", line 329, in __getitem__
[...]

Issue #323, opened 2022-09-24
[V2] Cache Library Properties


MiWeiss:
Most library properties create a list from the list of blocks. This can be a bit wasteful for large libraries if called often.

Thus, we should cache (and properly invalidate upon adding) any calls to such properties.

`functools.lru_cache` might be a solution.

MiWeiss:
overkill, and possibly unintended side-effects

PR #458, opened 2024-02-06
`Entry`: Dict-mimicking `in` operator.

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/454

### DIFF
--- a/bibtexparser/model.py
+++ b/bibtexparser/model.py
@@ -314,6 +314,10 @@ class Entry(Block):
         self._fields = [f for f in self._fields if f.key != key]
         return field
 
+    def __contains__(self, key: str) -> bool:
+        """Dict-mimicking ``in`` operator."""
+        return key in self.fields_dict
+
     def __getitem__(self, key: str) -> Any:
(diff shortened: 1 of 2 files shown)


bibtexparser/model.py
```diff
@@ -314,6 +314,10 @@ def pop_field(self, key: str) -> Optional[Field]:
         self._fields = [f for f in self._fields if f.key != key]
         return field
 
+    def __contains__(self, key: str) -> bool:
+        """Dict-mimicking ``in`` operator."""
+        return key in self.fields_dict
```
tdegeus, line 319: This does involve constructing a `dict`; however, I did not find a directer way. If there is, it would be good to optimize (now or in the future)

MiWeiss: I am not overly concerned about the dict creation. While it is slow, this can be mitigated by#323 in a future PR. 

For the next pre-release, I guess we can live with the way it is now (as barely anyone will use this yet), and then we should make sure to address #323 before releasing the first stable version.

Issue #455, opened 2024-02-03
Default `allow_parallel_execution=True`


tdegeus:
**Is your feature request related to a problem? Please describe.**
The idea of the middleware is great.

What I find a bit clumsy is the need to add a `True` argument for `allow_parallel_execution` if one writes and overload of `BlockMiddleware` without overloading the constructor. 

For example, this
```python
import bibtexparser
from bibtexparser.middlewares.middleware import BlockMiddleware

bibtex_str = """
@ARTICLE{Cesar2013,
author = {Jean César and T. W. J. de Geus},
title = {An amazing title},
year = {2013},
pages = {12--23},
volume = {12},
journal = {Nice Journal}
}
"""

class Foo(BlockMiddleware):
    def transform_entry(self, entry, *args, **kwargs):
        return entry

library = bibtexparser.parse_string(bibtex_str, append_middleware=[Foo()])
```
results in 
```python
Traceback (most recent call last):
  File "/Users/tdegeus/data/prog/forks/python-bibtexparser/t.py", line 19, in <module>
    library = bibtexparser.parse_string(bibtex_str, append_middleware=[Foo()])
                                                                       ^^^^^
TypeError: Middleware.__init__() missing 1 required positional argument: 'allow_parallel_execution'
```

I presume that most overloads will be threadsafe.

**Use cases**

Renders user-overloads more robust against future changes.

**Describe alternatives you've considered**
-

**Remaining Questions (Optional)**
Please tick all that apply:

- [ ] I would be willing to contribute a PR to fix this issue.
- [x] This issue is a blocker, I'd be grateful for an early fix.

tdegeus:
> 1. What's the reason for opening two distinct PRs? They appear to belong together. I do not particularly mind merging two PRs, but I am just wondering if I am missing a point. Is it because most Middleware already has the default set, but not the ones in [Adding missing default `allow_inplace_modification`. Simplifying examples #459](https://github.com/sciunto-org/python-bibtexparser/pull/459)?

The two concern different options. https://github.com/sciunto-org/python-bibtexparser/pull/459 just makes existing code consistent. Instead, https://github.com/sciunto-org/python-bibtexparser/pull/460 applies a default that was not there yet.

[...]

PR #460, opened 2024-02-06
`BlockMiddleware`: default `allow_parallel_execution=True`

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/455


MiWeiss (changes requested):
See comment in https://github.com/sciunto-org/python-bibtexparser/issues/455

MiWeiss (approved):
What an amazing addition to the docs! Thanks a lot @tdegeus

PR #462, opened 2024-02-07
Cleanup docstring

tdegeus:


MiWeiss (approved):
Thanks. We should probably define a default docstyle somewhere. 

I am afraid I haven't been fully consistent there. While I slightly prefer the `Google` style as it was before, there's already some `reST` at other places and as you also fixed some other things along the way I am happy to merge this.

PR #473, opened 2024-02-19
:sparkles: Add NormalizeFieldKeys middleware

### LINKED ISSUE #467: Normalize field keys (to lowercase)
**Describe the bug**
I have several .bib files that contain (mixed) field keys that are either in lowercase or start with a capital letter, such as "Author" and "Title". No other tooling complains about this.

SeparateCoAuthors does not work and I cannot uniformy access the fields using e.g. `entry['title']`

A normalization to lowercase of the field keys was conducted in v1.

Maybe this can be fixed using a middleware? I would be really grateful!

**Reproducing**

Version: e3757c13abf2784bda612464843ab30256317e6c

Code:
```python

#!/usr/bin/python

import bibtexparser
import bibtexparser.middlewares as m

layers = [
	m.LatexDecodingMiddleware(),
	m.MonthIntMiddleware(True), # Months should be represented as int (0-12)
	m.SeparateCoAuthors(True), # Co-authors should be separated as list of strings
	m.SplitNameParts(True), # Individual Names should be split into first, von, last, jr parts
	m.MergeNameParts("last", True) # Individual Names should be merged oto Last, First...
]

bib_database = bibtexparser.parse_file('data/Survey.bib', append_middleware=layers)
for entry in bib_database.entries:
	print(entry['title']);
```

Bibtex:
```
@InCollection{Name2006,
  Title                    = {A Title},
  Author                   = {Name, First and Name, Second},
  Booktitle                = {ITS},
  Publisher                = {Some publisher},
  Year                     = {2006},
  Pages                    = {61--70}
}
```

Please tick all that apply:

- [ ] I would be willing to contribute a PR to fix this issue.
- [ ] This issue is a blocker, I'd be grateful for an early fix.

Technologicat:
Here's an attempt at #467, which see.

Fixes #467


bibtexparser/middlewares/fieldkeys.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self, allow_inplace_modification: bool = True):
+        super().__init__(allow_inplace_modification=allow_inplace_modification,
+                         allow_parallel_execution=True)
+
+    def transform_entry(self, entry: Entry, library: "Library") -> Union[Block, Collection[Block], None]:
+        seen_normalized_keys: Set[str] = set()
+        new_fields_dict: Dict[str, Field] = {}
+        for field in entry.fields:
+            normalized_key: str = field.key.lower()
+            if normalized_key in seen_normalized_keys:
```
tdegeus, line 27: Since most entries will not have duplicate keys: is it cheaper to first check if the set is smaller than the list, and only then figure out which entries cause conflict?

bibtexparser/middlewares/fieldkeys.py
```diff
@@ -0,0 +1,37 @@
...
+                # TODO: Log the full entry, too, to help the user find where the failure occurred?
```
tdegeus, line 28: I guess only want to have the citation key (e.g. `smith2022`)

tests/middleware_tests/test_fieldkeys.py
```diff
@@ -0,0 +1,113 @@
...
+@inproceedings{jones2023,
+  Author    = "Jones, R.",
+  Title     = "A Test Conf. Paper",
+  Booktitle = "Proc. of the Intl. Test Conf.",
+  Year      = "2023",
+  Month     = 8
+}
+"""
+
+def test_normalize_lowercase():
```
tdegeus, line 57: Isn't the easiest check to ensure that there are no changes to `test_bibtex_string_lowercasekeys` is `NormalizeFieldKeys` is applied?

tests/middleware_tests/test_fieldkeys.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+
+    assert "author" in new_library.entries_dict["smith2022"]
+    assert new_library.entries_dict["smith2022"]["author"] == "Smith, J."
+    assert "author" in new_library.entries_dict["doe2021"]
+    assert new_library.entries_dict["doe2021"]["author"] == "Doe, J."
+    assert "author" in new_library.entries_dict["jones2023"]
+    assert new_library.entries_dict["jones2023"]["author"] == "Jones, R."
+
+def test_normalize_capitalized():
```
tdegeus, line 86: Here you could just check that `NormalizeFieldKeys` reduces `test_bibtex_string_capitalizedkeys` to `test_bibtex_string_lowercasekeys`

tests/middleware_tests/test_fieldkeys.py
Technologicat: I basically modeled the tests after `test_month.py`, because I'm not that familiar with `bibtexparser` yet. :)

The original string is preprocessed through a `Splitter`, which returns a `Library`. Then feeding the library (not the original string!) through the middleware returns a new `Library`. Libraries do not have an `__eq__` method, so comparing them will compare the `id`, thus failing the equality test.

I suppose if we export both libraries back into a string (to normalize whitespace) and compare those strings, this approach would be possible. But that relies on the exporter working correctly (i.e. this test unit is then no longer independent).

What do you think, shall I change it?

tests/middleware_tests/test_fieldkeys.py
Technologicat: Good idea. Same practical issue as above, so this depends on the answer to that.

bibtexparser/middlewares/fieldkeys.py
Technologicat: True.

It's a bit complicated to do that way, though, because our `transform_entry` method modifies each entry in-place as it encounters them. The deepcopy for the `allow_inplace_modification=False` mode takes place inside the middleware infrastructure, so also in that case, we mutate the copy we got as input. Thus we always lose the original unnormalized keys during the loop.

OTOH, I agree that performance is a concern, since nowadays big data is a thing, and one would expect libraries like `bibtexparser` to gracefully scale to at least into the 10k...100k range, so it would be nice to solve this.

We could dump a copy of the original unnormalized keys into another container, which we can then compare against the normalized key set after we are done. But then instead of one `if` switch per entry, we must run an `append` or something. All the extra allocations will cause pressure on the GC. Can we expect that solution to be any faster?

tests/middleware_tests/test_fieldkeys.py
tdegeus: I would be as simple/short as possible: parse back and forth to a string. However, you do make a fair point. @MiWeiss do you have a preference?

MiWeiss: > I suppose if we export both libraries back into a string (to normalize whitespace) and compare those strings, this approach would be possible. But that relies on the exporter working correctly (i.e. this test unit is then no longer independent).

Exactly, I likewise don't like that too much for the same reason. I would refrain from using the writer in any unit tests not directly related to bibtex writing.

Note: For the same reason, you *could* have avoided starting with a BibTeX string, but directly constructed your test data as Entry-objects. That way, the test would be really atomic. But on this end, I do not worry enough to make you re-write everything :-)

That's just my five cents though, as I was explicitly asked for it :-) If you decide to go either way, I'll be ok with it. In the end, @tdegeus is the reviewer of this PR...

tdegeus: No big preference. However, if you have the stamina, I would prefer the simplest based on `Entry`.

bibtexparser/middlewares/fieldkeys.py
tdegeus: I see. It appears that you are saying: I can check cheaply on uniqueness, but to have the error message as below, I would need a copy of fields before changing. 

So we have options:
* A more basic error message with just `entry.key`
* A constructor option to choose between basic and advanced error messages
* Keep things as they are

What is your preference? I could live with all

Technologicat: Yes, you nailed it. :)

My preference is almost always to go with maximally helpful error messages. It's much better UX for finding the error if the message explicitly says where and what it is. This is something the Python community generally does outstandingly well, and I want to respect that.

But my other preference is to keep things as simple as reasonably possible, following the Zen of Python. In this particular case, is the better UX worth the clarity and maintenance costs of the extra logic to support that UX?

I suppose the tradeoff depends on how long we expect a typical BibTeX entry to be. If it's just a few one-liners - e.g. author/year/title/journal/volume/pages/doi - a duplicate key error should be easy to spot. However, if the entry includes the full abstract, and a long bunch of extra metadata fields from some online service, well, then...

I'll mull over this overnight, and post a solution tomorrow.

tests/middleware_tests/test_fieldkeys.py
Technologicat: I'll see what I can do. This test unit is not that many lines of code...

bibtexparser/middlewares/fieldkeys.py
Technologicat: I've thought about this now.

I think it makes the most sense to keep this part as-is. The `if` is just one check per iteration. The alternative - if we want a helpful error message - is also one operation per iteration, to collect the original key names into a container. But in most cases, that container will never be read (only written to), causing unnecessary GC pressure.

Having a mode switch, on the other hand, increases the complexity. If squeezing out absolute maximal performance is really important, then having two separate implementations is justified, but otherwise I think it just complicates things.

If performance on large datasets becomes an issue, the simplest solution is to simplify the error message (making it less helpful).

I think that in the average application, the performance bottleneck will be somewhere else anyway, i.e. in what one does with the data `bibtexparser` reads in. For example, I'm running an NLP pipeline, semantic embeddings, and UMAP, which (even using GPU acceleration where available) takes ~2 orders of magnitude longer than the BibTeX parse step.

OTOH, use cases where BibTeX parsing makes up the majority of the runtime cost do also exist, such as a reference database format converter. In such cases maximal performance would be useful.

I suppose this comes down to which category of use cases we expect to be more common. My current bet is on applications with nontrivial processing, but please correct if I'm wrong. :)

tests/middleware_tests/test_fieldkeys.py
Technologicat: Reimplemented tests using `Entry`.

This doesn't make them shorter, but at least makes the test unit more independent. :)

Technologicat:
Also, noticed that `black` was complaining, so reformatted both `fieldkeys.py` and `test_fieldkeys.py` using it.

tests/middleware_tests/test_fieldkeys.py
Technologicat: The solution we ended up with (based on `Entry`) doesn't allow such a short comparison, unless we implement `__eq__` for `Library`. Perhaps too far beyond the scope of this PR?

bibtexparser/middlewares/fieldkeys.py
tdegeus: I'm fine with keeping what you suggested. We can always change if needed. Maybe a small comment in the code could be left?

Technologicat: Comment added. I'm not sure how to make it small, though. :)

tdegeus:
Thanks @Technologicat !! Since I had a few minutes to spare, I offered some simplifying suggestions by adding to commits. Please don't hesitate to criticize / complement / revert.

tdegeus (approved):
Good to go from my side!

Technologicat:
4b328ba: Ok, nice shortened comment. Thanks.
6ad724f: Ok.
9eed680: Ok.

71ee1de: Nice shortening of the test unit. Thanks!

Regarding this commit, some minor comments:

Contrary to the docstring, the function `test_normalize_fieldkeys` is actually testing with *lowercase* keys. I think the docstring should be fixed. This is important, because as the old maxim goes, *"if code and comments disagree, both are probably wrong"*.

Personally, I'd also avoid the name `"foo{i}"`. A semantically meaningful name, such as `"entry{i}"`, would increase clarity by making the intent explicit. Similarly, in `test_normalize_fieldkeys_force_last`, the bare `"foo"` could be e.g. `"dummyvalue"` to make it clear at a glance what it stands for.

(Using meaningful names exclusively is a habit I picked up from Racket docs ~half a decade ago. The standard metasyntactic names were fine in the 1970s, but we know better now. :) )

tdegeus:
Sounds good! Would you like to make the changes?

Technologicat:
Ugh, a merge conflict. Fixed. And some `flake8` warnings fixed, too.

Should be fine by me now.

tests/middleware_tests/test_fieldkeys.py
Technologicat: Test unit shortened by @tdegeus. Marking conversation as resolved.

tdegeus:
Thanks a million!

Technologicat:
Thanks for the merge!

Issue #464, opened 2024-02-07
Apply style using pre-commit


tdegeus:
To expedite applying formatting, one could consider `pre-commit` instead of manual configuring the CI. 

https://github.com/sciunto-org/python-bibtexparser/blob/57bdc8effe0d6fe418b577cf9ea0b75911a219a5/.github/workflows/linting_and_unittests.yml#L22

See e.g.

https://github.com/tdegeus/texplain/blob/main/.pre-commit-config.yaml
https://github.com/tdegeus/texplain/blob/main/.github/workflows/pre-commit.yml

MiWeiss:
Well, I am not a big fan of pre-commit. I found it to be slow and buggy. But then again the last time I gave it a try must be years ago. Thus, feel free to open a PR: No one is forced to use pre-commit locally, but it sure makes stuff nicer.

PR #474, opened 2024-02-23
:rotating_light: Apply automatic formatting

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/464


setup.py
```diff
@@ -47,11 +47,6 @@ def load_readme():
             "pytest-cov",  # Code coverage
             "jupyter",  # For runnable examples
         ],
-        "lint": [
-            "black==23.3.0",
```
MiWeiss, line 51: where are these defined for people not using pre-commit now?

MiWeiss (changes requested):
Thanks @tdegeus 

Besides questions marked inline, one more remark: There's now a ton of rules to be enforced. Maybe a bit more than I would have chosen, but that's not necessarily a problem ;-) Instead, I worry that this makes the use of pre-commit essentially mandatory. This however must not happen - the hurdle for anyone to contribute should be as low as possible.

I won't object to the strict rules (there's much good to them, probably, after all), but I think we should provide simple rules to what everyone can execute locally to make their code compatible, where possible. (Something that would have been `pip install -e .[lint] && black bibtexparser tests && isort bibtexparser tests --profile black` or alike). A simple instruction in the contribution guideline on how to enable pre-commit (or at least a link to the corresponding docs) would probably not hurt either - again, just to make it as simple as possible for everyone.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,61 @@
...
+  rev: v2.3.0
+  hooks:
+  - id: autoflake
+    args: [--in-place, --remove-unused-variable]
+- repo: https://github.com/pycqa/isort
+  rev: 5.13.2
+  hooks:
+  - id: isort
+    name: isort
+    args: ["--force-single-line", "--line-length=100", "--profile=black"]
```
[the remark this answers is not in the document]
tdegeus: No real argument. I guess it is because I was coming from https://github.com/asottile/reorder_python_imports . I like the `--force-single-line` for readability, though. I also think that a line-length of approx. 100 works well on modern screens.

setup.py
tdegeus: We could keep both, but I like the one-thing-one-place principle. To keep both, it would be nice if there would be an automatic way to keep config in sync

tdegeus:
* I would say that the key choice is to use `black`, that was and is the formatting rule.
* The minor change I proposed is for `isort`, which is less key, and I'm open to negotiate.
* The extra 'rule' is to truncate trailing whitespace everywhere, which is, I think, what most editors do, and therefore what results in the cleanest PRs.
* The rest is mostly checks, such as small 'mistakes' that have been removed in `rst` and f-strings. 

I could completely support supporting a second way to run these tools locally. It would be nice though to enforce keeping their settings in sync (which would be assured if we decide to go back to the default for `isort`, so I can easily support that). Running with pre-commit it's not hard either, though:
```
python -m pip install pre-commit
pre-commit run --all-files
```
and we should definitely document.
The advantage, I find, is that one does not have to keep track of which tools to run.

setup.py
MiWeiss: So do I, if possible. Your example posted below will I fine, I guess :-)

```
python -m pip install pre-commit
pre-commit run --all-files
```

MiWeiss:
Thanks for your comment. Sounds all good. I'll give the "formal" ( :smile: ) approval after the documentation update has been committed.

tdegeus:
Done. If you have comments / suggestions please feel free!

tdegeus:
Thanks, sorry I overlooked that. Fixed.

Issue #475, opened 2024-03-10
split_multiple_persons_names() bug


nschloe:
`split_multiple_persons_names()` doesn't give the correct splitting if there's a one-character middle name, e.g., `Doe, John A`.

MWE:

```python
from bibtexparser.middlewares.names import split_multiple_persons_names

# okay:
string = "Doe, John and Doe, Jane"
print(split_multiple_persons_names(string))

# okay:
string = "Doe, John A. and Doe, Jane"
print(split_multiple_persons_names(string))

# fail:
string = "Doe, John A and Doe, Jane"
print(split_multiple_persons_names(string))
```
```python
['Doe, John', 'Doe, Jane']
['Doe, John A.', 'Doe, Jane']
['Doe, John A and Doe, Jane']
```

This is with v2.0.0b7.

tdegeus:
Thanks for reporting! This is indeed a bug. Should be fixed in https://github.com/sciunto-org/python-bibtexparser/pull/476

PR #476, opened 2024-03-14
:bug: fix `split_multiple_persons_names` with single-char middlenames

tdegeus:
Fixes https://github.com/sciunto-org/python-bibtexparser/issues/475


MiWeiss (approved):
thanks

PR #489, opened 2024-07-30
💥[BREAKING CHANGE] - Rename `write_file` parameters

cinntamani:
Correct naming of arguments and docstring.


MiWeiss (approved):

MiWeiss:
Nice. I like it, and I am glad we do this breaking change while we're still in beta :-)

Thanks

PR #491, opened 2024-10-10
👌 IMPROVE: Support for keeping abbreviations at enclosing values.

lartpang:
@drakeguan @jagd @mgorny @sjpfenninger 

In this commit, I add a new argument `keep_abbr_string` into `AddEnclosingMiddleware` to keep the abbreviation (`@string`) when enclosing the value and not resolving the string.

### DIFF
--- a/tests/middleware_tests/test_enclosing.py
+++ b/tests/middleware_tests/test_enclosing.py
@@ -197,6 +197,100 @@ def _figure_out_added_enclosing(changed_value, value):
     return used_enclosing
 
 
+@pytest.mark.parametrize("metadata_resolving", ["", "journal"])
+@pytest.mark.parametrize("metadata_enclosing", ["{", '"', "no-enclosing", None])
+@pytest.mark.parametrize("default_enclosing", ["{", '"'])
+@pytest.mark.parametrize("enclose_ints", [True, False], ids=["enclose_ints", "no_enclose_ints"])
+@pytest.mark.parametrize(
+    "keep_abbr_string", [True, False], ids=["keep_abbr_string", "no_keep_abbr_string"]
+)
+@pytest.mark.parametrize("reuse_previous_enclosing", [True, False], ids=["reuse", "no_reuse"])
+@pytest.mark.parametrize(
+    "value",
+    [
+        # value, is a abbreviation?
+        ("IEEE_T_PAMI", True),
+        ('IEEE_T_PAMI # "ieee tpami"', True),
+        ('IEEE_T_PAMI" # ieee tpami', False),
+        ('IEEE_T-PAMI # "ieee tpami"', False),
+        ('IEEE_T-PAMI # "ieee # tpami"', False),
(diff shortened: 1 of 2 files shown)


tests/middleware_tests/test_enclosing.py
```diff
@@ -197,6 +197,100 @@ def _figure_out_added_enclosing(changed_value, value):
...
+        ('IEEE_T_PAMI # "ieee tpami"', True),
+        ('IEEE_T_PAMI" # ieee tpami', False),
+        ('IEEE_T-PAMI # "ieee tpami"', False),
+        ('IEEE_T-PAMI # "ieee # tpami"', False),
+        ('IEEE T-PAMI # "ieee tpami"', False),
+    ],
+)
+@pytest.mark.parametrize("inplace", [True, False], ids=["inplace", "not_inplace"])
+def test_addition_of_enclosing_on_entry_with_abbr(
+    value: tuple,
```
MiWeiss, line 222:
```suggestion
    value: tuple[str,bool],
```

MiWeiss:
Hi @lartpang 

Thanks a lot for your PR.

This is certainly a feature we want to support. However, there's two major comments to address:

- In principle, I try to keep a match between "Parse" and "Write" Middleware, where the latter (partially) reverts the first. In your case: The stings were dereferenced in `ResolveStringReferencesMiddleware`, thus they should be put back in a `ResolveStringReferencesMiddleware` (in interpolate.py). I see how that's not very straight forward (as its very connected to the enclosings) - but please try to consider moving your solution in that direction. Then, if you absolutely can't find a nice solution, just say so. I'll have a look and re-evaluate (I'm a bit short on time atm).
- Make sure the test suite covers end-to-end cases, i.e., cases where the `ResolveStringReferencesMiddleware` works together with your middleware changes. That's important to protect from regressions, as the two are very much tied together.

I'll check for minor things after we settled on these, but I don't expect any hickups after that :-)

Thanks again

lartpang:
@MiWeiss Thanks for your review.

> In principle, I try to keep a match between "Parse" and "Write" Middleware, where the latter (partially) reverts the first. In your case: The strings were dereferenced in `ResolveStringReferencesMiddleware`, thus they should be put back in a `ResolveStringReferencesMiddleware` (in interpolate.py). I see how that's not very straight forward (as its very connected to the enclosings) - but please try to consider moving your solution in that direction. Then, if you absolutely can't find a nice solution, just say so. I'll have a look and re-evaluate (I'm a bit short on time atm).

`ResolveStringReferencesMiddleware` is used to resolve string objects in bibtex, which seems to conflict with the feature of retaining string objects in this commit: 
Because if `ResolveStringReferencesMiddleware` is used, it explicitly indicates that string objects should be resolved, so there would be no need to emphasize "retaining string objects" additionally.
The requirement to "retain string objects" only exists when enclosing values (maybe), which seems difficult to decouple from the operation of `AddEnclosingMiddleware`.

PR #505, opened 2026-01-14
♻️ Replace assertions with explicit exceptions

MiWeiss:
Replace assert statements with explicit if + raise ValueError in _cast_to_duplicate() to ensure checks run even with Python -O flag.


MiWeiss:
change by claude

PR #523, opened 2026-06-11
:bug: Fix misleading 'Unknown block type' warning for failed blocks and add missing DuplicateFieldKeyBlock.entry (#520)

### LINKED ISSUE #520: What is bibtexparser.model.DuplicateFieldKeyBlock ?
I'm getting

`Unknown block type <class 'bibtexparser.model.DuplicateFieldKeyBlock'>`

when I apply  `bibtexparser.parse_file ` to some of my .bib files.  But then it doesn't seem possible to dump the fields to find the source of the problem.  

And `bibtexparser.model.DuplicateFieldKeyBlock` is not documented.

MiWeiss:
Fixes #520

## What was wrong

As reported in #520, parsing a bib file containing an entry with duplicate field keys produced the confusing warning

```
Unknown block type <class 'bibtexparser.model.DuplicateFieldKeyBlock'>
```

and offered no documented way to inspect the problem. Three distinct issues were behind this:

1. **Misleading warning:** `BlockMiddleware.transform_block` had no branch for `ParsingFailedBlock` and its subtypes, so any block middleware (including `RemoveEnclosingMiddleware` from the *default* parse stack) fell through to the "Unknown block type" warning for every failed block — even though these are well-known model types that are intentionally passed through.
2. **Wrong error message:** the `DuplicateFieldKeyBlock` error message told users the entry is "available as `failed_block.entry`", but no such attribute exists — the entry is available as `failed_block.ignore_error_block`. This is the "another problem shadowing correct behavior" suspected in the issue discussion.
3. **Missing documentation:** the failed-block types were not part of the API docs, and the quickstart's error-checking section referred to a "corresponding section of the docs" that didn't exist.

## Changes

- `BlockMiddleware` now handles `ParsingFailedBlock` explicitly via a new overridable `transform_failed_block` hook, which by default keeps failed blocks unchanged (and silent). Custom middlewares can override it if they want to process failed blocks.
- Fixed the `DuplicateFieldKeyBlock` error message to reference the existing `ignore_error_block` attribute (consistent with the other `ParsingFailedBlock` subtypes), instead of the non-existent `.entry`.
- Documented `ParsingFailedBlock`, `MiddlewareErrorBlock`, `DuplicateBlockKeyBlock` and `DuplicateFieldKeyBlock` in the API docs, added `Library.failed_blocks` to the `Library` docs, and expanded the quickstart error-checking section with the available attributes (`error`, `start_line`, `raw`, `ignore_error_block`, ...).
- Bumped two pre-commit hook pins (`language-formatters-pre-commit-hooks`, `docstr_coverage`) whose pinned versions crash in current CI environments with `ModuleNotFoundError: No module named 'pkg_resources'`; their latest releases dropped the `pkg_resources` dependency.
- Tests: failed blocks pass through `BlockMiddleware` unchanged without any warning; `DuplicateFieldKeyBlock.duplicate_keys` and the corrected error message are asserted in the splitter tests.

## Before / after

```python
import bibtexparser
lib = bibtexparser.parse_string("""
@article{Smith2020,
  title = {A title},
  title = {A duplicate title},
}
""")
```

Before: logs `Unknown block type <class 'bibtexparser.model.DuplicateFieldKeyBlock'>`; the error message points to `failed_block.entry`, which raises `AttributeError`.

After: no warning; the error message points to `failed_block.ignore_error_block`, which returns the entry including both `title` fields, and `lib.failed_blocks[0].duplicate_keys == {"title"}`.

Full test suite passes (2475 passed, 12 skipped), pre-commit hooks pass.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
   - id: check-toml
   - id: debug-statements
 - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
-  rev: v2.12.0
+  rev: v2.16.0
   hooks:
   - id: pretty-format-yaml
     args: [--preserve-quotes, --autofix, --indent, '2']
@@ -55,7 +55,7 @@ repos:
   hooks:
   - id: setup-cfg-fmt
 - repo: https://github.com/HunterMcGushion/docstr_coverage
-  rev: v2.3.0
+  rev: v2.3.2
   hooks:
   - id: docstr-coverage
     args: ["bibtexparser"]
--- a/bibtexparser/middlewares/middleware.py
+++ b/bibtexparser/middlewares/middleware.py
@@ -9,6 +9,7 @@ from bibtexparser.model import Block
 from bibtexparser.model import Entry
 from bibtexparser.model import ExplicitComment
 from bibtexparser.model import ImplicitComment
+from bibtexparser.model import ParsingFailedBlock
 from bibtexparser.model import Preamble
 from bibtexparser.model import String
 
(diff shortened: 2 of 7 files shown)


MiWeiss (commented):
reviewed

Issue #478, opened 2024-04-13
Question: sorting blocks by specific field values


oborin1:
**Is your feature request related to a problem? Please describe.**
I want to sort my bibtex blocks using values of a specific field, e.g., year. Is it already possible with SortBlocksByTypeAndKeyMiddleware()?

**Describe the solution you'd like**
Unsorted bibtex file --> Sorted bibtex file with blocks arranged in ascending order, e.g., by year.

**Use cases**
1. Merge transliterated and originally Latin bibliographies having author-year citation mode as part of translation.
2. Sort an untidy bibliography.

**Describe alternatives you've considered**
None.

**Remaining Questions (Optional)**
- [x ] I would be willing to contribute a PR to fix this issue.
- [ ] This issue is a blocker, I'd be grateful for an early fix.

MiWeiss:
Thanks for your question. Which version of `bibtexparser` do you use? `v1.0.0x` or `v1.4.x`?

oborin1:
Thank you for your prompt answer. I'm currently migrating from v1.4.x to v2.0.0x.

MiWeiss:
Great!

The `SortBlocksByTypeAndKeyMiddleware` is not exactly what you want, as it implements a predefined order based on the block types (entry, string, ...) and the keys of the blocks.

Instead you'd want to create your own custom middleware extending `LibraryMiddleware` using your desired logic. Here, you may want to take the [SortBlocksByTypeAndKeyMiddleware](https://github.com/sciunto-org/python-bibtexparser/blob/main/bibtexparser/middlewares/sorting_blocks.py) Implementation as an example.

In the long run, we'll certainly want to expose a simple way to achieve this (e.g. by exposing a dedicated abstract class for sorting, or a class which takes a comparator as constuctor argument), but as of now, this is not yet supported.

MiWeiss:
I created https://github.com/sciunto-org/python-bibtexparser/issues/480 to track the progress of the above mentioned improvements.

MiWeiss:
closed as answered and inactive

PR #525, opened 2026-06-11
✨ Add SortBlocksMiddleware for custom block sorting

### LINKED ISSUE #480: ✨ Allow simple way to create custom block sorting
We'd want to facilitate users sorting blocks according to their needs. 

They easiest and most general way to do this would probably be to create a new abstract class `SortBlocksMiddleware` with the following properties:

- The abstract class should have just one unimplemented method, and it should have the signature `compare(block1, block2) -> int`, following the standard convention that any returned value > 0 indicates that block2 should come anywhere after block1, any value <0 the opposit and a return value of 0 sais that any order of the two should be fine. 
- The docstring of the `compare` method should list the standard conventions for such a method (the conditions above, but also that it's transitive, pure, ...)
- Our own `SortBlocksByTypeAndKeyMiddleware` should extend/implement `SortBlocksMiddleware`
- The documentation should provide a simple MWE

Location of the code: https://github.com/sciunto-org/python-bibtexparser/blob/main/bibtexparser/middlewares/sorting_blocks.py

I am currently not working on this, so feel free to comment below if you want to take over.

MiWeiss:
﻿Closes #480 (see also #478, where this need came up).

## What this does

Adds a generic `SortBlocksMiddleware` that sorts the blocks of a library by a user-provided sort-key function, working like the `key` argument of Python's built-in `sorted`:

```python
from bibtexparser.middlewares import SortBlocksMiddleware
from bibtexparser.model import Entry

def by_year(block):
    # Tuple keys make mixed block types comparable:
    # non-entries and year-less entries on top, the rest sorted by year
    if isinstance(block, Entry) and year in block:
        return (1, int(block[year]))
    return (0, 0)

middleware = SortBlocksMiddleware(key=by_year)
```

`reverse=True` and `preserve_comments_on_top` are supported; comment-preservation reuses the existing `_BlockJunk` grouping, so custom sorts keep comments attached to their entry for free.

## Design deviation from the issue body

#480 proposed an abstract class with an unimplemented `compare(block1, block2) -> int`. Following the later discussion in the issue (thanks @mirhahn), this PR implements a **key-function** API instead:

- `key=` is the native Python 3 sorting idiom (`cmp` was removed from `list.sort` in Python 3); writing a correct, transitive comparator is harder than `lambda e: (e[year], e.key)`.
- Comparator users get a free escape hatch via `functools.cmp_to_key` (documented in the class docstring); there is no conversion in the other direction.
- The existing middlewares (`SortBlocksByTypeAndKeyMiddleware`, both field-sorting middlewares) already sort via key functions internally — a comparator-based abstract class would have required converting working key functions into comparators.

Also, the class is **concrete** (key passed as constructor argument, mirroring `SortFieldsCustomMiddleware`) rather than abstract, so one-off sorts need no subclass; subclassing for reusable named middlewares still works.

## Backwards compatibility

`SortBlocksByTypeAndKeyMiddleware` is now a thin subclass of `SortBlocksMiddleware` supplying its `(type_index, block.key)` tuple key. Behavior is unchanged; the pre-existing tests pass unmodified.

## Review aid

The new unit tests each parse a small literal BibTeX string and assert the resulting order, so the functionality is directly visible: sort by year, reverse, hierarchical `(year, author)` keys, sort stability, mixed block types, and both comment-handling modes. Docs (`customize.rst`) gained an MWE under the Sorting section plus a note pointing custom-sorting users to the new class.

### DIFF
--- a/bibtexparser/middlewares/__init__.py
+++ b/bibtexparser/middlewares/__init__.py
@@ -15,6 +15,7 @@ from bibtexparser.middlewares.names import NameParts
 from bibtexparser.middlewares.names import SeparateCoAuthors
 from bibtexparser.middlewares.names import SplitNameParts
 from bibtexparser.middlewares.sorting_blocks import SortBlocksByTypeAndKeyMiddleware
+from bibtexparser.middlewares.sorting_blocks import SortBlocksMiddleware
 from bibtexparser.middlewares.sorting_entry_fields import SortFieldsAlphabeticallyMiddleware
 from bibtexparser.middlewares.sorting_entry_fields import SortFieldsCustomMiddleware
 
--- a/bibtexparser/middlewares/sorting_blocks.py
+++ b/bibtexparser/middlewares/sorting_blocks.py
@@ -1,6 +1,8 @@
 from copy import deepcopy
 from dataclasses import dataclass
 from dataclasses import field
+from typing import Any
+from typing import Callable
 from typing import List
 from typing import Tuple
 from typing import Type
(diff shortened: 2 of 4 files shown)


MiWeiss:
reviewed

Issue #400, opened 2023-09-20
Write of files with duplicate blocks fails


claell:
**Describe the bug**
Writing a library with duplicate blocks leads to:

```
WARNING:root:Unknown block type <class 'bibtexparser.model.DuplicateBlockKeyBlock'>
Traceback (most recent call last):
  File "n:\Gruppenplatte\Veröffentlichungen\2023\Systems Engineering Journal (INCOSE)\Scopus retrieval\test.py", line 24, in <module>
    bibtexparser.write_file("my_new_file.bib", bib_library)
  File "C:\Users\Ellsel\AppData\Local\Programs\Python\Python311\Lib\site-packages\bibtexparser\entrypoint.py", line 136, in write_file
    bibtex_str = write_string(
                 ^^^^^^^^^^^^^
  File "C:\Users\Ellsel\AppData\Local\Programs\Python\Python311\Lib\site-packages\bibtexparser\entrypoint.py", line 168, in write_string
    return write(library, bibtex_format=bibtex_format)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ellsel\AppData\Local\Programs\Python\Python311\Lib\site-packages\bibtexparser\writer.py", line 106, in write
    string_block_pieces = _treat_block(bibtex_format, block)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ellsel\AppData\Local\Programs\Python\Python311\Lib\site-packages\bibtexparser\writer.py", line 127, in _treat_block
    string_block_pieces = _treat_failed_block(block, bibtex_format)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Ellsel\AppData\Local\Programs\Python\Python311\Lib\site-packages\bibtexparser\writer.py", line 71, in _treat_failed_block
    lines = len(block.raw.splitlines())
                ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'splitlines'
[...]

PR #526, opened 2026-06-12
:bug: Raise descriptive error when writing failed blocks without raw bibtex

MiWeiss:
## Context

Fixes the crash reported in #400: writing a library containing a `ParsingFailedBlock` whose `raw` is `None` (e.g. a `DuplicateBlockKeyBlock` produced by adding two programmatically created entries with the same key) failed with an opaque `AttributeError: 'NoneType' object has no attribute 'splitlines'`.

## Change

The rule "failed block ⇒ write raw bibtex with a warning comment" is kept unchanged. When there is no raw bibtex, the writer no longer crashes but raises a single, actionable `ValueError`:

- `write()` pre-scans the library and lists **all** offending blocks (with their errors) in one exception, so users can fix everything in one pass instead of replaying write-fail-fix.
- The message points to `library.failed_blocks` and suggests removing the blocks or fixing and re-adding `block.ignore_error_block`.
- `_treat_failed_block` keeps a backstop raise for direct callers.

Refusing to write (rather than silently falling back to serializing `ignore_error_block`) was discussed in #400: it loses no data, keeps the writer rule simple, and surfaces the duplicate-key bug to the user instead of shipping it downstream.

A follow-up PR proposes flipping the `Library.add` default to `fail_on_duplicate_key=True`, which would surface this class of error at the `add()` call site already.


MiWeiss:
reviewed

PR #527, opened 2026-06-12
🦖 Make Library.add fail on duplicate keys by default

MiWeiss:
## Context

Companion to #526, addressing the root cause of #400: `library.add(entry)` silently converted a programmatically created entry into a `DuplicateBlockKeyBlock` when its key already existed. Users only discovered this much later (e.g. at write time), far from the `add()` call that caused it.

## Change

**Breaking (pre-2.0-final):** `Library.add` and the `Library` constructor now default to `fail_on_duplicate_key=True` and raise a `ValueError` on duplicate keys.

The failing `add` is **atomic**: duplicates (including duplicates within the same `add` call) are detected by a pre-check, so the exception is raised before any mutation and the library is left untouched. Passing `fail_on_duplicate_key=False` restores the silent-replacement behavior (duplicates become `DuplicateBlockKeyBlock`s, exposed via `library.failed_blocks`).

This also makes `add` consistent with `Library.replace`, which has always defaulted to `fail_on_duplicate_key=True`.

**Parsing behavior is unchanged:** the splitter and the middleware library reconstruction (`BlockMiddleware.transform`, `SortBlocksByTypeAndKeyMiddleware`) explicitly pass `fail_on_duplicate_key=False`, so bibtex files with duplicate keys still parse leniently into `library.failed_blocks` (covered by the existing `test_handles_duplicate_entries`).

Since v2 is still in beta (v2.0.0b9), this seems like the last cheap moment to flip the default.

### DIFF
--- a/bibtexparser/library.py
+++ b/bibtexparser/library.py
@@ -17,23 +17,30 @@ from .model import String
 class Library:
     """A collection of parsed bibtex blocks."""
 
-    def __init__(self, blocks: Union[List[Block], None] = None):
+    def __init__(
+        self,
+        blocks: Union[List[Block], None] = None,
+        fail_on_duplicate_key: bool = True,
+    ):
         self._blocks = []
         self._entries_by_key = dict()
         self._strings_by_key = dict()
         if blocks is not None:
-            self.add(blocks)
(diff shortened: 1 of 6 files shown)


MiWeiss:
reviewed

PR #528, opened 2026-06-12
✨ Honor enclosing demands on fields and strings when writing (#447)

### LINKED ISSUE #447: [question] How to write months as abbreviations?
Instead of `month = {1}`, I would like to see `month = jan` in my bibtex outputs. I see that `write_string` and `write_file` both have some middleware argument, but something like

```python
write_string(
    library,
    bibtex_format=bibtex_format,
    prepend_middleware=[MonthAbbreviationMiddleware(True)],
)
```

only gives

```
month = {jan}
```

(with brackets).

Any hints?

MiWeiss:
Closes #447

## What this PR does

Adds an explicit, always-honored *enclosing demand* to field and string values, and uses it to make month macros (and other bibtex string references) survive writing with the default stack.

### 1. `Field.enclosing` / `String.enclosing` (new model attribute)
- Allowed values: `'{'`, `'"'`, `'no-enclosing'`, or `None` (default = writer middleware decides).
- `AddEnclosingMiddleware` honors a non-None demand with precedence over `reuse_previous_enclosing`, the int-rule, and `default_enclosing`.
- Assigning a new `value` resets the demand to `None`, so a stale demand can never be applied to a changed value (and the demand is naturally "consumed" once the enclosed value is written).

### 2. Month middlewares set the demand (the actual #447 use-case)
`MonthAbbreviationMiddleware` and `MonthIntMiddleware` now demand `no-enclosing` for the macros/ints they produce. With the default unparse stack:

```python
library = bibtexparser.parse_string("@article{a, month = {1}}")
bibtexparser.write_string(library, prepend_middleware=[MonthAbbreviationMiddleware()])
# now contains `month = jan` (string reference), previously `month = {jan}` (literal)
```

`MonthLongStringMiddleware` deliberately sets no demand (full month names are literals).

### 3. Parse→write roundtrip no longer corrupts references (latent bug)
Previously, a plain `parse_string` → `write_string` roundtrip turned `month = jan` into `month = {jan}` and `pages = intro # outro` into `{intro # outro}` — silently changing semantics. `RemoveEnclosingMiddleware` now sets the `no-enclosing` demand on unenclosed, non-numeric values. Bare integers (`year = 2019`) remain governed by `enclose_integers`, as before.

**Note:** This changes default write output for unenclosed non-numeric inputs (verbatim instead of brace-normalized). I'd argue the old behavior was a bug since it changed semantics, but it's in a separate commit (7e37177) for easy discussion/removal.

### Drive-by fixes
- `AddEnclosingMiddleware._enclose` no longer crashes on `int` values when `enclose_integers=False` (`value.isdigit()` → `AttributeError`).
- Month middlewares' unchanged-month paths returned the `Field` object instead of its value (assigning the field to itself as `field.value`).

## Design discussion
Alternatives considered (marker `str` subclass, parser-metadata convention, month special-casing in the writer) — the explicit attribute was chosen because it is discoverable, expresses any enclosing preference (not just "none"), is robust against field-key renames/duplicates, and fails safe via the value-setter reset.

### DIFF
--- a/bibtexparser/middlewares/enclosing.py
+++ b/bibtexparser/middlewares/enclosing.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from typing import Tuple
 from typing import Union
 
@@ -78,6 +79,17 @@ class AddEnclosingMiddleware(BlockMiddleware):
 
     This middleware adds enclosing characters to a field value.
     It is useful when the field value is enclosed in braces or quotes.
+
+    The enclosing to add is determined with the following precedence:
+
+    1. An enclosing demanded on the field or string itself
+       (i.e., a non-None ``field.enclosing``, e.g. set by a month middleware)
+       is always honored.
+    2. If ``reuse_previous_enclosing`` is True, the enclosing removed by the
+       ``RemoveEnclosingMiddleware`` (if any) is used.
+    3. If the value is an integer in a common int field
+       and ``enclose_integers`` is False, no enclosing is added.
+    4. Otherwise, the ``default_enclosing`` is used.
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


MiWeiss:
reviewed

PR #555, opened 2026-06-12
✏️ Fix typos and docstring mismatches

MiWeiss:
- `_val_intent_string` → `_val_indent_string`
- `BibtexFormat.indent` docstring claimed the default is a space (it is a tab); `value_column` docstring claimed String blocks are aligned (they are not)
- `_BlockJunk` → `_BlockChunk` (incl. user-facing error message)
- double assignment in `Block.__init__`
- missing spaces in concatenated warning/error messages; grammar fixes in exceptions

Fixes #542

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #542, opened 2026-06-12
Assorted typos and docstring mismatches


MiWeiss:
Small issues, collected in one place:

- `BibtexFormat.indent` docstring says "Default: single space", but the default is a tab (`bibtexparser/writer.py:135`, `:143`)
- `BibtexFormat.value_column` docstring claims it affects String blocks, but `_treat_string` never aligns (`writer.py:41-48`, `:154`)
- `_val_intent_string` — typo for "indent" (`writer.py:35`)
- `_BlockJunk` is presumably meant to be "chunk"; "junk" also appears in a user-facing error message (`bibtexparser/middlewares/sorting_blocks.py:24`, `:36`)
- `Block.__init__` assigns `_parser_metadata` twice (`bibtexparser/model.py:34-36`)
- Missing spaces in concatenated message strings: `bibtexparser/middlewares/interpolate.py:49-52`, `bibtexparser/middlewares/enclosing.py:133`, `:166`, `bibtexparser/exceptions.py:53`
- "Exception where a invalid bibtex file" (`exceptions.py:22`)

MiWeiss:
All items here were fixed in #555 (merged) and the String alignment in #558 (merged); verified against current main (ca6b9bf).

PR #558, opened 2026-06-12
✨ Align @string values according to BibtexFormat.value_column

MiWeiss:
The `value_column` docstring has claimed since the original writer draft (#356) that String blocks are aligned, but `_treat_string` never applied any alignment. This implements it, with the semantics decided explicitly:

- **Integer column**: string values use the same padding rule as entry fields, measured after the `@string{` prefix — entry and string values thus each align among themselves (not at the same absolute column, since the prefixes differ).
- **`"auto"`**: the longest key now considers *both* entry field keys and string keys, so a single consistent column is used. Note this can shift entry alignment in libraries with long string keys.

Default (`value_column = 0`) output is unchanged.

**Stacked on #555** (uses the `_val_indent_string` rename) — merge #555 first; this PR's diff will then shrink to the writer/test changes only.

Follow-up to the discussion on #555 / #542.

### DIFF
--- a/bibtexparser/exceptions.py
+++ b/bibtexparser/exceptions.py
@@ -19,7 +19,7 @@ class ParsingException(Exception):
 
 
 class BlockAbortedException(ParsingException):
-    """Exception where a invalid bibtex file led to an aborted block."""
+    """Exception where an invalid bibtex file led to an aborted block."""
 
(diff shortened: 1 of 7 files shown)


MiWeiss:
reviewed

Issue #452, opened 2024-01-25
Enclosing middleware fails for certain content


claell:
**Describe the bug**
https://bibtexparser.readthedocs.io/en/main/customize.html#default-parse-stack says that there is Enclosing middleware used, likely to make sure no illegal characters exist in fields.

**Reproducing**

Version: `2.0.0b2`

Note the abstract of the following excerpt (`The 396$\{2}$ chip is`):

Bibtex:
```
@Article{Konstadinidis2009,
  author = {Konstadinidis, Georgios K. and Tremblay, Marc and Chaudhry, Shailender and Rashid, Mamun and Lai, Peter F. and Otaguro, Yukio and Orginos, Yannis and Parampalli, Sudhendra and Steigerwald, Mark and Gundala, Shriram and Pyapali, Rambabu and Rarick, Leonard D. and Elkin, Ilyas and Ge, Yuefei and Parulkar, Ishwar},
  title = {Architecture and physical implementation of a third generation 65 nm, 16 Core, 32 thread chip-multithreading SPARC processor},
  date = {2009-01-01},
  doi = {10.1109/JSSC.2008.2007144},
  abstract = {This third-generation Chip-Multithreading (CMT) SPARC processor consists of 16 cores with shared memory architecture and supports a total of 32 main threads plus 32 scout threads. It is targeted for high-performance servers, and is optimized for both single- and multi-threaded applications. The 396$\{2}$ chip is fabricated in an 11 metal layer 65-nm CMOS process and operates at a nominal frequency of 2.3 GHz, consuming a maximum power of 250 W at 1.2 V. This paper provides an overview of the architectural highlights and describes the physical implementation challenges and solutions including circuit innovations in memory arrays, register files, and floating-point hardware that boost the performance and circuit robustness with low area overhead. © 2006 IEEE.},
  author_keywords = {Arrays; Chip multi-threading (CMT); Clocking; Computer architecture; Execute ahead; Hardware scout; Microprocessor; Multi-core; Multi-threaded; Power management; Register files; SerDes; SPARC architecture; Transactional memory}
}
```

**Workaround**
Manually looking for the problematic character

**Remaining Questions (Optional)**
Please tick all that apply:

- [ ] I would be willing to contribute a PR to fix this issue.
- [ ] This issue is a blocker, I'd be grateful for an early fix.

claell:
This gives me an error when parsing the file (written from bibtexparser):
```
[...]

Issue #567, opened 2026-07-16
Trailing backslashes break name round trips


claell:
**Describe the bug**

Trailing backslashes break the intended inverse relationship between
`NameParts.merge_last_name_first` and `parse_single_name_into_parts`.

Two related cases are involved:

* Parsing a name whose final token ends in a backslash appends that backslash
  twice because the end-of-input branch does not stop further handling of the
  same character.
* `merge_last_name_first` escapes the final section even though no comma follows
  it, changing a terminal backslash unnecessarily.

The existing inverse-property test skips strict and non-strict examples with
odd trailing backslashes, concealing the round-trip failure.

**Reproducing**

Version: current `main` (`08e370a`)

Code:

```python
from bibtexparser.middlewares.names import NameParts
from bibtexparser.middlewares.names import parse_single_name_into_parts

parts = NameParts(
    first=["Christina", "A.", "L.\\"],
    von=[],
    last=["Thiele"],
    jr=[],
)
merged = parts.merge_last_name_first
assert parse_single_name_into_parts(merged) == parts
```

The assertion fails on current `main`.

**Expected behavior**

Last-name-first serialization and reparsing should preserve the `NameParts`
value in both strict and non-strict modes, including terminal backslashes. Only
sections followed by a comma need delimiter-protecting escaping.

**Workaround**

Avoid terminal backslashes in name parts or normalize affected names manually.

**Remaining Questions (Optional)**

* [x] I would be willing to contribute a PR to fix this issue.
* [ ] This issue is a blocker, I'd be grateful for an early fix.

claell:
Draft PR #573 contains the focused parser/serializer fix and enables the previously skipped strict and non-strict inverse-property cases.

MiWeiss:
Hi @claell — I'm posting this same note on all of your July 16 issues and PRs (#567–#597), so apologies for the form-letter feel.

I'm closing this. The batch — 13 issues, 18 PRs, ~3,000 lines, opened within a two-hour window with spec-style language and forward-references to PR numbers that didn't exist yet — reads as AI-generated rather than something you hit and verified by hand. Reviewing it properly would cost me more time than just fixing the parser myself.

If this is a real bug you've actually hit: reopen it with a concrete repro and a small, human-verified fix, and I'll review it. Any nontrivial design or API choice should be discussed in the issue first, before code is written. Otherwise, please disclose and verify AI-assisted contributions before submitting them in future.

claell:
[...]

PR #572, opened 2026-07-16
Preserve fields after escaped brace sequences

claell:
## Summary

* Keep field brace-depth accounting stable when an opening brace is escaped.
* Preserve every field following the escaped-brace sequence.
* Add a public parse/write/reparse regression so a shallow successful parse
  cannot conceal later-field loss.

Follow-up to #452. The earlier report was closed after discussion of enclosing
middleware, but a stricter reproduction shows that current `main` silently
drops all fields following the reported abstract sequence.

## Validation

* Complete enclosing regression matrix under warnings-as-errors: 917 passed, 10
  skipped.
* Complete upstream suite: 2,577 passed, 12 skipped, with the four existing
  deprecation warnings in `tests/test_entrypoint.py`.
* `git diff --check`: passed.

## Review note

exercised over a long period in production. Because the splitter previously
returned a normal entry while losing later fields, careful human review of the
brace-state change and regression fixture is requested.

## AI assistance

maintainer review.

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -197,6 +197,12 @@ class Splitter:
                 return next_mark.start()
             # Check for end of entry:
             elif next_mark.group(0) == "}" and not _is_escaped():
+                # A malformed escaped-brace sequence can reduce the tracked depth too early.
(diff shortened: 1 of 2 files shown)


MiWeiss:

PR #573, opened 2026-07-16
Preserve trailing backslashes in name round trips

### LINKED ISSUE #567: Trailing backslashes break name round trips

claell:
## Summary

* Stop processing a terminal backslash twice after the name parser reaches end
  of input.
* Escape only last-name-first sections that are followed by a comma.
* Enable the existing strict and non-strict inverse-property cases instead of
  skipping names with odd trailing backslashes.

Fixes #567.

## Validation

* Complete name middleware suite under warnings-as-errors: 689 passed.
* Complete upstream suite: 2,578 passed, 10 skipped, with the four existing
  deprecation warnings in `tests/test_entrypoint.py`. The two removed skips are
  the now-passing strict and non-strict inverse cases.
* `git diff --check`: passed.

## Review note

exercised over a long period in production. The broad existing name fixture
matrix now passes without the targeted skip, but careful human review of TeX
escaping semantics is requested.

## AI assistance

maintainer review.

### DIFF
--- a/bibtexparser/middlewares/names.py
+++ b/bibtexparser/middlewares/names.py
@@ -143,8 +143,14 @@ class NameParts:
         last = " ".join(self.last) if self.last else None
         jr = " ".join(self.jr) if self.jr else None
 
-        von_last = " ".join(name for name in [von, last] if name)
-        return ", ".join(escape_last_slash(name) for name in [von_last, jr, first] if name)
+        sections = [
+            name for name in [" ".join(name for name in [von, last] if name), jr, first] if name
+        ]
+
+        # Only sections followed by a comma need protection from a trailing backslash.
+        # Escaping the final section changes its value even though no delimiter follows it.
+        escaped_sections = [escape_last_slash(section) for section in sections[:-1]]
+        return ", ".join([*escaped_sections, sections[-1]])
 
 
 class SplitNameParts(_NameTransformerMiddleware):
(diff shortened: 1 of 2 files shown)


MiWeiss:

PR #574, opened 2026-07-16
Require exact special command types

### LINKED ISSUE #568: Custom entry types matching special-command prefixes are misclassified
**Describe the bug**

The splitter classifies block types with `startswith`, so valid custom entry
types whose names begin with `comment`, `preamble`, or `string` are routed to a
special-command parser instead of the ordinary entry parser.

For example, `@commentary`, `@preamble_study`, and `@string_theory` do not become
entries even though only the exact case-insensitive names `comment`, `preamble`,
and `string` should select those special block forms.

**Reproducing**

Version: current `main` (`08e370a`)

Code:

```python
from bibtexparser import parse_string

for entry_type in ("commentary", "preamble_study", "string_theory"):
    library = parse_string(
        f"@{entry_type}{{key, title = {{Custom entry type}}}}"
    )
    assert [(entry.entry_type, entry.key) for entry in library.entries] == [
        (entry_type, "key")
    ]
```

The assertions fail on current `main` because the blocks are treated as special
commands rather than custom entries.

**Expected behavior**

Special-command dispatch should use an exact case-insensitive type comparison.
Longer custom type names should remain ordinary entries.

**Workaround**

Rename custom entry types so they do not begin with a special command name.

* [x] I would be willing to contribute a PR to fix this issue.
* [ ] This issue is a blocker, I'd be grateful for an early fix.

claell:
## Summary

* Dispatch `comment`, `preamble`, and `string` as special commands only on an
  exact case-insensitive type match.
* Preserve longer custom types such as `commentary`, `preamble_study`, and
  `string_theory` as ordinary entries.

Fixes #568.

## Validation

* Focused custom-entry regression matrix under warnings-as-errors: 3 passed.
* Complete upstream suite: 2,579 passed, 12 skipped, with the four existing
  deprecation warnings in `tests/test_entrypoint.py`.
* `git diff --check`: passed.

## Review note

exercised over a long period in production. Although the dispatch change is
small and covered across all three prefixes, careful human review is requested.

## AI assistance

maintainer review.

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -300,6 +300,7 @@ class Splitter:
             m_val = m.group(0).lower()
 
             if m_val.startswith("@"):
+                block_type = m_val[1:].strip()
                 # Clean up previous block implicit_comment
                 implicit_comment = self._end_implicit_comment(m.start())
                 if implicit_comment is not None:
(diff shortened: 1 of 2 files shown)


MiWeiss:

PR #578, opened 2026-07-16
Isolate and strengthen field-sorting tests

### LINKED ISSUE #577: Isolate and strengthen field-sorting tests
## Problem

The field-sorting middleware tests reuse one module-level mutable `Entry`. Because in-place middleware cases reorder that shared object, later tests can inherit earlier mutations and assert an order that depends on test execution history rather than the documented behavior.

The suite also leaves several useful contracts implicit:

* fields omitted from a custom order retain their original relative order;
* custom ordering works through the public writer without mutating the source library when in-place modification is disabled; and
* ambiguous custom orders with duplicate keys, including case-only duplicates under case-insensitive matching, fail early.

## Proposed coverage

Construct a fresh entry for each test, correct the stable-order assertion, and add focused public-writer and duplicate-order cases. This should be a test-only change because current production behavior already satisfies the contracts.

claell:
## Summary

* Isolate in-place field-sorting cases with fresh entry fixtures.
* Verify stable ordering for fields omitted from a custom order.
* Exercise custom sorting through the public writer without source mutation.
* Reject ambiguous custom orders, including case-only duplicates.

Closes #577.

## Validation

* `pytest -p no:cacheprovider -W error tests/middleware_tests/test_sorting_entry_fields.py`: 8 passed.
* Complete upstream test suite: 2,579 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This PR changes only tests. In particular, it replaces an execution-order-dependent expected value with the stable ordering produced from a fresh input. Maintainer review should confirm that unspecified fields retaining their original relative order is the intended contract.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with test isolation, behavior analysis, branch preparation, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/tests/middleware_tests/test_sorting_entry_fields.py
+++ b/tests/middleware_tests/test_sorting_entry_fields.py
@@ -1,27 +1,32 @@
 import pytest
 
 from bibtexparser import Library
+from bibtexparser import write_string
 from bibtexparser.middlewares.sorting_entry_fields import SortFieldsAlphabeticallyMiddleware
 from bibtexparser.middlewares.sorting_entry_fields import SortFieldsCustomMiddleware
 from bibtexparser.model import Entry
 from bibtexparser.model import Field
 
-TEST_ENTRY = Entry(
-    "article",
-    "Cesar2013",
-    [
-        Field("author", "Cesar, J.", 1),
-        Field("title", "A title", 2),
-        Field("note", "A note", 3),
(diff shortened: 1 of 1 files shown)


MiWeiss:

PR #580, opened 2026-07-16
Exercise enclosure-like values without skipped tests

### LINKED ISSUE #579: Exercise enclosure-like values without skipped tests
## Problem

Ten enclosure middleware cases are currently skipped when a value itself starts and ends with braces or quotes. Those cases are ambiguous only if the test tries to infer whether a wrapper was added from the final string; the middleware's expected behavior can still be asserted directly.

The skips leave gaps in both directions:

* removal should record and strip enclosure-like boundary characters consistently, even when the test did not add an outer wrapper; and
* addition with `no-enclosing` should leave an enclosure-like value exactly unchanged rather than being skipped because it resembles wrapped output.

## Proposed coverage

Replace the skip helper with explicit expected-value and metadata assertions for removal, and compare exact output for `no-enclosing` addition. This turns the complete enclosure matrix into executable regression coverage without changing production behavior.

claell:
## Summary

* Replace enclosure-like value skips with explicit assertions.
* Verify removal values and `removed_enclosing` metadata across the full matrix.
* Assert exact output when addition is configured with `no-enclosing`.

Closes #579.

## Validation

* `pytest -p no:cacheprovider -W error tests/middleware_tests/test_enclosing.py`: 926 passed, with no skips.
* Complete upstream test suite: 2,586 passed, 2 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This is a test-only change. The key judgment is that values beginning and ending with enclosure characters remain valid cases to test: removal interprets those boundary characters as enclosure, while addition configured with `no-enclosing` adds nothing and should be checked by exact equality.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with test-matrix analysis, branch isolation, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/tests/middleware_tests/test_enclosing.py
+++ b/tests/middleware_tests/test_enclosing.py
@@ -16,13 +16,6 @@ from tests.resources import EDGE_CASE_VALUES
 from tests.resources import ENCLOSINGS
 
 
-def _skip_pseudo_enclosing_value(value: str):
-    starts_and_ends_in_brackets = value.startswith("{") and value.endswith("}")
-    starts_and_ends_in_quotes = value.startswith('"') and value.endswith('"')
-    if starts_and_ends_in_quotes or starts_and_ends_in_brackets:
-        pytest.skip("No enclosing to remove")
-
-
 @pytest.mark.parametrize("enclosing", ENCLOSINGS + [pytest.param("{0}", id="no_enclosing")])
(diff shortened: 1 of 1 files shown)


MiWeiss:

PR #571, opened 2026-07-16
Report unsupported parenthesis-delimited blocks

claell:
## Summary

* Detect standard parenthesis-delimited entries, strings, preambles, and
  comments instead of classifying them as implicit comments.
* Retain their exact source in explicit `ParsingFailedBlock` objects while full
  parenthesis parsing remains unsupported.
* Recover later supported blocks on the same or following line, including after
  an unclosed parenthesis block.

Fixes #533 by taking the issue's fail-closed interim option. This does not claim
to implement full parenthesis-delimited parsing.

## Validation

* Focused splitter detection and recovery suite under warnings-as-errors: 32
  passed.
* Complete upstream suite: 2,585 passed, 12 skipped, with the four existing
  deprecation warnings in `tests/test_entrypoint.py`.
* `git diff --check`: passed.
* The project pre-commit executable was not available in the isolated local
  environment; GitHub CI should independently run the configured hooks.

## Review note

This change was developed and validated in a concentrated session rather than
exercised over a long period in production. The recovery scanner handles nested
parentheses, braces, quotes, and subsequent blocks, but careful human review of
malformed-input edge cases is particularly important.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high
reasoning effort. Codex assisted with analysis, implementation, branch
isolation, and test execution. Automated validation is not a substitute for
maintainer review.

PR #584, opened 2026-07-16
Recognize complete BibTeX block starts

### LINKED ISSUE #583: Recognize complete BibTeX block starts
## Problem

The splitter recognizes a block start with `@\w*` followed only by spaces or tabs before the opening delimiter. This is narrower than the BibTeX data-language structure in two important ways:

* line endings and other ordinary whitespace may separate an entry type from its outer `{` or `(` delimiter; and
* entry-type identifiers may contain valid non-word punctuation such as `-`, `:`, or `/`.

Candidates outside the regular expression are currently retained as implicit comments rather than becoming parsed blocks or explicit failures. That makes a structural input disappear from the bibliography model without any parse failure.

## Proposed behavior

Recognize the complete block-type token up to whitespace or a reserved BibTeX syntax character, locate the following standard outer delimiter across whitespace, and use the delimiter position rather than assuming it immediately follows the regex match.

Add a matrix requiring every recognized candidate to become either:

* a parsed block for the currently supported curly-brace form; or
* an explicit failed block for the currently unsupported parenthesis form.

This work builds on the explicit parenthesis-failure handling in #533 / PR #571 and exact special-command matching in #568 / PR #574. The implementation PR should retain those as separate prerequisite commits so the broader parser change remains reviewable.

claell:
## Summary

* Recognize ordinary BibTeX whitespace, including line endings, before outer block delimiters.
* Preserve punctuation permitted in entry-type identifiers instead of treating the candidate as an implicit comment.
* Require the structural matrix to produce either a parsed block or an explicit failed block; no candidate may disappear silently.

Closes #583.

This draft currently includes the exact prerequisite behavior from:

* PR #571, which retains unsupported parenthesis-delimited blocks as explicit failures; and
* PR #574, which requires exact matching for `@comment`, `@preamble`, and `@string`.

Those behaviors remain separate as the first two commits. The issue #583 implementation and its exhaustive visibility matrix are the final two commits. The branch is executable against current `main`; once the prerequisites merge, it should be rebased so their already-reviewed diffs disappear from this PR.

## Validation

* `pytest -p no:cacheprovider -W error tests/splitter_tests/test_splitter_block_start_detection.py`: 75 passed.
* Complete upstream suite: 2,628 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This changes the splitter's block-start grammar, so the draft should receive careful maintainer review despite the broad structural matrix. In particular, review should confirm the reserved-character set used to terminate entry-type identifiers and the decision to keep parenthesis syntax explicit but unsupported for now.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with parser analysis, dependency separation, conflict resolution, test design, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -376,6 +376,7 @@ class Splitter:
             m_val = m.group(0).lower()
 
             if m_val.startswith("@"):
+                block_type = m_val[1:].strip()
                 # Clean up previous block implicit_comment
                 implicit_comment = self._end_implicit_comment(m.start())
                 if implicit_comment is not None:
@@ -387,11 +388,11 @@ class Splitter:
                     # Start new block parsing
                     if self.bibstr[m.end()] == "(":
                         library.add(self._handle_parenthesis_block(m))
-                    elif m_val.startswith("@comment"):
+                    elif block_type == "comment":
                         library.add(self._handle_explicit_comment())
-                    elif m_val.startswith("@preamble"):
+                    elif block_type == "preamble":
                         library.add(self._handle_preamble())
-                    elif m_val.startswith("@string"):
+                    elif block_type == "string":
                         library.add(self._handle_string(m), fail_on_duplicate_key=False)
                     else:
                         library.add(self._handle_entry(m, m_val), fail_on_duplicate_key=False)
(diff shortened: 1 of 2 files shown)


MiWeiss:

PR #585, opened 2026-07-16
Preserve Biber comments within entries

### LINKED ISSUE #372: Support biber comments

claell:
## Summary

* Parse Biber-style percent-comment lines inside entries before field parsing can absorb them as data.
* Represent nested comments separately from active bibliography fields, with their source field-boundary positions.
* Preserve those comments during canonical writes, including after callers remove fields.
* Cover field-shaped comments, punctuation, CRLF input, top-level-comment compatibility, copying, and model invariants.

Closes #372.

## Default behavior

This is implemented in the structural parser rather than as opt-in middleware. A middleware cannot recover a comment that has already caused field parsing to fail or absorb subsequent syntax. The parser therefore recognizes a line whose first non-whitespace character is `%` inside an entry and retains it in `Entry.comments`; it does not expose the text as an active field.

The canonical writer preserves content and relative field boundaries, not byte-identical whitespace. Callers may inspect or edit the explicit `EntryComment` objects. No dialect flag is required merely to avoid losing Biber input.

## Validation

* Focused Biber-comment and affected model tests under warnings-as-errors: 50 passed.
* Complete upstream suite: 2,584 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This adds public model state to `Entry` and changes splitter marks, so careful API and parser review is requested. The branch includes the same small `_skip_marks_before` iterator helper independently needed by PR #571; whichever PR merges second should remove the duplicate during rebase.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with architecture analysis, dependency removal, edge-case testing, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/model.py
+++ b/bibtexparser/model.py
@@ -306,6 +306,47 @@ class Field:
         )
 
 
+class EntryComment:
+    """A Biber-style ``%`` comment positioned between fields of an entry.
+
+    ``field_index`` records how many fields preceded the comment in the parsed
+    source. It lets the canonical writer retain comments while keeping
+    ``Entry.fields`` restricted to active bibliography data.
+    """
+
+    def __init__(self, comment: str, field_index: int):
+        if field_index < 0:
+            raise ValueError("field_index must be non-negative")
+        self._comment = comment
+        self._field_index = field_index
+
+    @property
+    def comment(self) -> str:
+        """Content following the source line's leading ``%`` marker."""
+        return self._comment
+
+    @comment.setter
+    def comment(self, value: str):
+        self._comment = value
+
+    @property
+    def field_index(self) -> int:
+        """Number of active fields that preceded this comment when parsed."""
+        return self._field_index
+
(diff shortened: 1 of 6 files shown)


MiWeiss:

PR #566, opened 2026-07-16
Honor custom parsing failure comments

sapunyangkut:
## Summary

- Honor the active `BibtexFormat.parsing_failed_comment` template when writing failed blocks.
- Strengthen the existing regression test for both the default and a custom template.

Fixes #530

## Validation

Preserved exact-SHA evidence for this unchanged upstream base and identical patch:
- `pytest tests/test_writer.py::test_write_failed_block`: 1 passed
- `pytest tests/test_writer.py`: 25 passed
- configured pre-commit hooks on both changed files: passed

The current no-install checkout lacks `pylatexenc`, so its fresh focused rerun stopped during collection. No dependencies or dependency files were changed.

## AI assistance

This Draft PR was prepared with OpenAI Codex. Automated root-cause analysis, patching, and validation evidence were used; no human self-review is claimed. It remains a draft for maintainer and user review.

Issue #530, opened 2026-06-12
BibtexFormat.parsing_failed_comment is ignored by the writer


MiWeiss:
`BibtexFormat` exposes a settable `parsing_failed_comment` property (`bibtexparser/writer.py:201-207`), but `_treat_failed_block` uses the module constant `PARSING_FAILED_COMMENT` instead (`bibtexparser/writer.py:64-67`). Setting the property has no effect on output.

PR #589, opened 2026-07-16
Define an idempotent failed-block writing policy

### LINKED ISSUE #588: Define an idempotent failed-block writing policy
## Problem

`ParsingFailedBlock.raw` is the only retained representation of input the parser could not model. The writer currently prepends a generated warning comment to that source by default. On the next parse, the warning becomes an ordinary implicit comment; writing again adds another warning. Repeated read/write cycles therefore accumulate generated content even when the user changes nothing.

There is also no fail-closed writer mode for applications that must refuse export while any parse failure remains.

## Proposed policy

Add an explicit `BibtexFormat.failed_block_policy` with three values:

* `"preserve"` (proposed default): write retained raw failed blocks unchanged;
* `"annotate"`: prepend the configured diagnostic, while recognizing and replacing an identical existing annotation so repeated cycles are idempotent; and
* `"raise"`: refuse to write if the library contains any failed block, reporting all failures together.

An unknown policy must raise immediately. Failed blocks without raw source remain unwritable under every policy. There should be no silent discard mode.

This setting applies only to failed blocks. Normal parsed blocks must continue to serialize from their current model state because their cached raw source may be stale after middleware or user edits.

## Compatibility

Changing the default from annotation to preservation is intentional but user-visible. The former behavior remains available as `"annotate"`, and strict applications gain `"raise"`. The documentation should call out the change and the fact that canonical writing is not a byte-identical concrete-syntax editor for normal blocks.

The configurable annotation text is already tracked by #530 / PR #566. A standalone implementation of this proposal may carry that fix as a clearly separated prerequisite until it merges.

claell:
## Summary

* Preserve retained failed-block raw source unchanged by default.
* Offer explicit `annotate` and fail-closed `raise` policies.
* Keep opt-in annotations idempotent across repeated parse/write cycles.
* Reject unknown policies and continue rejecting failed blocks without raw source.
* Document why normal mutable blocks still serialize from model state rather than cached raw text.

Closes #588.

The first commit is the same focused configured-comment fix already represented by #530 / draft PR #566. It is included so the policy branch is executable against current `main`. The policy and annotation-idempotence changes are the second and third commits. After #566 merges, this branch should be rebased to remove the prerequisite diff.

## Validation

* Focused failed-block writer and round-trip tests under warnings-as-errors: 8 passed before the annotation follow-up, followed by the dedicated annotation-idempotence regression passing.
* Complete upstream suite: 2,583 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

The proposed default changes from generated annotation to exact raw preservation. This is a deliberate integrity choice, but it is user-visible and needs maintainer review. The previous behavior remains available explicitly as `annotate`; strict consumers gain `raise`. There is intentionally no discard policy.

The implementation also avoids adding a second terminal newline to failed raw source. Without that detail, even `annotate` changes the failure's line count and cannot recognize its own diagnostic on the next cycle.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with policy design, dependency isolation, regression analysis, newline edge-case repair, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/writer.py
+++ b/bibtexparser/writer.py
@@ -12,6 +12,7 @@ from .model import String
 
 VAL_SEP = " = "
 PARSING_FAILED_COMMENT = "% WARNING Parsing failed for the following {n} lines."
+FAILED_BLOCK_POLICIES = ("preserve", "annotate", "raise")
 
 
 def _treat_entry(block: Entry, bibtex_format) -> list[str]:
@@ -63,6 +64,12 @@ def _treat_expl_comment(block: ExplicitComment, bibtex_format: "BibtexFormat") -
 def _treat_failed_block(block: ParsingFailedBlock, bibtex_format: "BibtexFormat") -> list[str]:
     if block.raw is None:
         raise ValueError(_failed_blocks_without_raw_error([block]))
+    if bibtex_format.failed_block_policy == "preserve":
+        return [block.raw]
+    if bibtex_format.failed_block_policy == "raise":
+        # The complete-library check in `write` normally reports all failures at
+        # once. Keep this guard so direct internal use cannot bypass the policy.
+        raise ValueError(_failed_blocks_forbidden_error([block]))
     lines = len(block.raw.splitlines())
     parsing_failed_comment = bibtex_format.parsing_failed_comment.format(n=lines)
     return [parsing_failed_comment, "\n", block.raw, "\n"]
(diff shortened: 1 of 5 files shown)


MiWeiss:

PR #591, opened 2026-07-16
Make string reference expansion opt-in

### LINKED ISSUE #590: Make string reference expansion opt-in
## Problem

The default parse stack applies `ResolveStringReferencesMiddleware`. An entry such as `journal = journalName` therefore becomes the current literal value of `@string{journalName = ...}`. Writing the parsed library emits the expanded value and loses the relationship to the macro definition.

That is not merely formatting normalization: later edits to the `@string` definition no longer update the entry, and a parse/write cycle creates many duplicated literals. The original reference cannot be reconstructed reliably from the expanded text.

## Proposed default

Remove `ResolveStringReferencesMiddleware` from the default parse stack so references remain references. Keep the middleware public and document an explicit dereferencing profile:

```python
layers = [
    ResolveStringReferencesMiddleware(),
    RemoveEnclosingMiddleware(),
]
```

Ordering matters because interpolation must see the original expression before ordinary enclosure removal changes its representation.

Add public-entry-point tests proving both contracts:

* default parsing and writing retain `@string` linkage; and
* callers who explicitly request interpolation receive expanded values.

## Compatibility

This is a user-visible default change. Existing users who rely on immediate expanded values must add the middleware explicitly. The alternative is irreversible semantic loss during ordinary parse/write use, so preservation is the safer default for a bibliography codec.

Concatenated expressions and cyclic/unresolved references are related to #396 but are not prerequisites for deciding the default. The initial implementation should use basic references and remain independently reviewable.

claell:
## Summary

* Retain references to `@string` definitions in the default parse stack.
* Keep `ResolveStringReferencesMiddleware` available as an explicit dereferencing choice.
* Document and test the required middleware order before enclosure removal.
* Protect macro linkage through the public parse/write entry points.

Closes #590.

## Validation

* `pytest -p no:cacheprovider -W error tests/middleware_tests/test_interpolate.py`: 5 passed.
* Complete upstream suite: 2,578 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This is a user-visible default change: callers who rely on immediate literal values must add `ResolveStringReferencesMiddleware` explicitly. The reason for proposing it is that default expansion is irreversible during a write—macro linkage is semantic information, and the original reference cannot be reconstructed reliably from its expanded text.

The PR deliberately uses basic references and does not depend on the broader concatenation work discussed in #396.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with default-policy analysis, standalone test design, documentation, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/middlewares/parsestack.py
+++ b/bibtexparser/middlewares/parsestack.py
@@ -1,4 +1,3 @@
-from bibtexparser.middlewares import ResolveStringReferencesMiddleware
 from bibtexparser.middlewares.enclosing import AddEnclosingMiddleware
 from bibtexparser.middlewares.enclosing import RemoveEnclosingMiddleware
 
@@ -8,7 +7,6 @@ from .middleware import Middleware
 def default_parse_stack(allow_inplace_modification: bool = True) -> list[Middleware]:
     """The default parse stack to be applied after splitting, if not specified otherwise."""
     return [
-        ResolveStringReferencesMiddleware(allow_inplace_modification=allow_inplace_modification),
         RemoveEnclosingMiddleware(allow_inplace_modification=allow_inplace_modification),
     ]
 
--- a/docs/source/customize.rst
+++ b/docs/source/customize.rst
@@ -48,8 +48,21 @@ part of the expected functionality for most users.
 
 Currently, the default parse stack consists of the following layers:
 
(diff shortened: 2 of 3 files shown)


MiWeiss:

PR #593, opened 2026-07-16
Reuse parsed value enclosures by default

### LINKED ISSUE #592: Reuse parsed value enclosures by default
## Problem

`RemoveEnclosingMiddleware` already records whether each parsed value used braces, quotes, or no enclosure. The default writer nevertheless configures `AddEnclosingMiddleware(reuse_previous_enclosing=False)`, so an unchanged parse/write cycle converts quoted and bare numeric values to braces.

That creates unrelated diffs and means editing a field's value also changes its established enclosure style. This is avoidable layout mutation because the parser already has the necessary metadata.

## Proposed default

Set `reuse_previous_enclosing=True` in the default unparse stack:

* parsed quoted values remain quoted;
* parsed braced values remain braced;
* parsed bare integers remain bare;
* explicitly demanded bare macro/expression values remain bare; and
* new programmatic fields without parser metadata continue using braces.

## Safety for edited bare values

Reusing `no-enclosing` metadata requires one guard. A bare integer may later be edited to ordinary text while retaining the original metadata. Emitting that text bare could create invalid BibTeX. When `no-enclosing` comes only from parser metadata (rather than an explicit `field.enclosing` demand), reuse it only while the current value is still numeric; otherwise fall back to the configured default enclosure.

Tests should cover unchanged quote/brace/bare values, edits to quoted and braced fields, the edited-bare fallback, and values created without metadata.

## Compatibility

This is a formatting-default change, not a semantic transformation. Callers who prefer uniform braces can continue to supply `AddEnclosingMiddleware(reuse_previous_enclosing=False, ...)` explicitly. Preserving source enclosure choices is the less surprising default for ordinary read/edit/write use.

claell:
## Summary

* Reuse parsed quote, brace, and safe bare enclosure styles in the default writer.
* Preserve established quote/brace style when users edit field values.
* Fall back to braces if a formerly bare integer is edited to ordinary text.
* Keep braces as the default for new programmatic values without parser metadata.
* Extend the enclosure matrix so stale `no-enclosing` metadata cannot authorize unsafe text.

Closes #592.

## Validation

* Complete enclosure suite under warnings-as-errors: 928 passed, 2 skipped.
* Complete upstream suite: 2,588 passed, 4 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

The safety rule also turns eight formerly skipped matrix combinations into executable assertions because ordinary text with stale bare metadata now has an unambiguous enclosed result.

## Review note

This changes default formatting but preserves semantics and reduces unrelated diffs. Callers that require uniform braces can still configure `AddEnclosingMiddleware(reuse_previous_enclosing=False, ...)` explicitly.

The edited-bare guard is important: blindly applying `no-enclosing` parser metadata after a value changes can emit invalid BibTeX. An explicit `field.enclosing = "no-enclosing"` demand still takes precedence for callers intentionally supplying a valid bare expression.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with default-policy analysis, unsafe-edit discovery, matrix refinement, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/middlewares/enclosing.py
+++ b/bibtexparser/middlewares/enclosing.py
@@ -148,7 +148,12 @@ class AddEnclosingMiddleware(BlockMiddleware):
         if demanded_enclosing is not None:
             enclosing = demanded_enclosing
         elif self._reuse_previous_enclosing and metadata_enclosing is not None:
-            enclosing = metadata_enclosing
+            # Parser metadata describes the source value, not necessarily its
+            # current edited value. Reusing a bare integer's metadata for new
+            # ordinary text would emit invalid BibTeX, so fall back to the
+            # configured default unless the bare representation remains safe.
+            if metadata_enclosing != "no-enclosing" or str(value).isdigit():
+                enclosing = metadata_enclosing
         elif apply_int_rule and not self._enclose_integers and str(value).isdigit():
             return str(value)
 
--- a/bibtexparser/middlewares/parsestack.py
+++ b/bibtexparser/middlewares/parsestack.py
@@ -19,7 +19,7 @@ def default_unparse_stack(allow_inplace_modification: bool = False) -> list[Midd
         AddEnclosingMiddleware(
             allow_inplace_modification=allow_inplace_modification,
             default_enclosing="{",
-            reuse_previous_enclosing=False,
+            reuse_previous_enclosing=True,
             enclose_integers=True,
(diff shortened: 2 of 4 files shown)


MiWeiss:

PR #595, opened 2026-07-16
Define a non-lossy semantic round-trip contract

### LINKED ISSUE #594: Define a non-lossy semantic round-trip contract
## Problem

The suite has many focused parser and writer tests but no explicit end-to-end definition of what ordinary `parse_string` → `write_string` → `parse_string` use must preserve. Without a central contract, a change can keep producing syntactically valid output while silently losing or reordering meaningful bibliography information.

Byte identity is too strict for the current canonical writer: line endings, indentation, spacing, optional commas, and some enclosure formatting may normalize. Short-circuiting unchanged files by returning cached raw source would also be the wrong test because it would not validate the model produced by the parser or the writer's serialization of edits.

## Proposed semantic contract

Define a reusable semantic inventory over the parsed model and require it to survive a complete parse/write/reparse path. It should include, in source order:

* block kind and block order;
* entry type, citation key, field order, field-key spelling, and field values;
* string definitions and preambles;
* explicit and implicit comment content and position; and
* retained failed-block raw source.

The contract should fail loudly when a new model block type is added without deciding how it participates.

Add separate cases for:

* preserving block and field order unless sorting middleware is requested;
* permitting documented layout normalization while semantic inventory remains equal;
* normalizing whitespace-only input to empty output;
* serializing edited and newly added fields correctly while unrelated records remain semantically unchanged; and
* retaining failed input through the model, not through a raw-file bypass.

## Scope boundary

This is a semantic round-trip contract for the existing canonical writer, not a byte-preserving concrete-syntax editing mode. A future exact mode could preserve untouched token spans and rewrite only edited nodes, but that requires a concrete-syntax model and should be proposed separately.

The failed-input portion depends on the raw-preserving default proposed in #588 / PR #589. Default macro-linkage and enclosure-style choices remain separately reviewable in #590 / PR #591 and #592 / PR #593 rather than being bundled into this contract.

claell:
## Summary

* Define an ordered semantic inventory for entries, strings, preambles, comments, and failed blocks.
* Require full parse/write/reparse paths to preserve that inventory without a cached-raw bypass.
* Protect block and field order, comment position, and unrelated records during edits.
* Permit documented canonical cleanup such as line-ending, whitespace, and enclosure normalization.
* Fail explicitly if a future model block type is not classified by the contract.

Closes #594.

The first three commits are the exact failed-block behavior from PR #589, which is required for retained failures to survive without generated annotation blocks. The semantic contract is the final test-only commit. Once #589 merges, this branch should be rebased so reviewers see only the contract diff.

The default macro-linkage and enclosure-style proposals in PRs #591 and #593 are intentionally independent. This contract does not bundle those product decisions.

## Validation

* `pytest -p no:cacheprovider -W error tests/test_roundtrip_contract.py`: 6 passed.
* Complete upstream suite: 2,589 passed, 12 skipped, with the four existing deprecation warnings in `tests/test_entrypoint.py`.

## Review note

This is deliberately not a byte-identity promise. The writer remains canonical and may normalize documented layout. The protected invariant is that all modeled bibliography meaning survives in order, including retained failed raw source, and that edits serialize through the actual model and writer.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with contract design, dependency reduction, fixture adaptation, and validation. Automated validation is not a substitute for maintainer review.

### DIFF
--- a/bibtexparser/writer.py
+++ b/bibtexparser/writer.py
@@ -12,6 +12,7 @@ from .model import String
 
 VAL_SEP = " = "
 PARSING_FAILED_COMMENT = "% WARNING Parsing failed for the following {n} lines."
+FAILED_BLOCK_POLICIES = ("preserve", "annotate", "raise")
 
 
 def _treat_entry(block: Entry, bibtex_format) -> list[str]:
@@ -63,6 +64,12 @@ def _treat_expl_comment(block: ExplicitComment, bibtex_format: "BibtexFormat") -
 def _treat_failed_block(block: ParsingFailedBlock, bibtex_format: "BibtexFormat") -> list[str]:
     if block.raw is None:
         raise ValueError(_failed_blocks_without_raw_error([block]))
+    if bibtex_format.failed_block_policy == "preserve":
+        return [block.raw]
+    if bibtex_format.failed_block_policy == "raise":
+        # The complete-library check in `write` normally reports all failures at
+        # once. Keep this guard so direct internal use cannot bypass the policy.
+        raise ValueError(_failed_blocks_forbidden_error([block]))
     lines = len(block.raw.splitlines())
     parsing_failed_comment = bibtex_format.parsing_failed_comment.format(n=lines)
     return [parsing_failed_comment, "\n", block.raw, "\n"]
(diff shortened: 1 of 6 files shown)


MiWeiss:

PR #582, opened 2026-07-16
Capture expected deprecation warnings in entry-point tests

claell:
## Summary

* Capture the expected `DeprecationWarning` in legacy/new parameter conflict tests.
* Retain the existing `ValueError` and message assertions.
* Make the complete suite pass with warnings treated as errors.

Closes #581.

## Validation

* `pytest -p no:cacheprovider -W error`: 2,576 passed, 12 skipped.
* Complete pre-commit hook suite: passed.

## Review note

This is a test-only change. It makes warnings already emitted by the intended conflict path explicit rather than suppressing warnings globally.

## AI assistance

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with warning analysis, branch isolation, and validation. Automated validation is not a substitute for maintainer review.

PR #597, opened 2026-07-16
Add read-only compatibility preflight

### LINKED ISSUE #596: Add a read-only compatibility preflight
## Motivation

Users may have long-lived or unusual `.bib` files that no finite fixture suite covers. Given past silent-loss cases, a user should be able to ask whether the installed parser's default public path appears safe for a specific file before trusting it with edits.

This cannot prove that the parser interpreted every construct correctly: parsing and reparsing with the same implementation is not an independent oracle. It can still detect important failure classes and provide a concrete, read-only confidence report.

## Proposed API

Add opt-in `check_string(...)` and `check_file(...)` APIs returning an immutable, JSON-serializable report. Required checks should include:

* **source coverage:** ordered raw block spans account for every non-whitespace source token;
* **explicit parse success:** no retained failed blocks;
* **semantic round trip:** default write/reparse preserves the protected ordered model inventory;
* **canonical stability:** a second cycle reaches the same canonical output; and
* **source identity:** report separately whether default writing would leave the source bytes/text unchanged or merely preserve semantics while normalizing layout.

If explicit parse failures already make the file incompatible, expensive or misleading downstream checks may be marked `not run`. Diagnostics must use stable codes and avoid embedding bibliography content or paths.

The protected inventory should share the same implementation as semantic round-trip regression tests so the runtime report and suite cannot drift. New model block types must require an explicit contract decision.

## CLI

Provide an installed command and `python -m bibtexparser` entry point:

```text
bibtexparser check FILE [--encoding ...] [--json]
```

Exit statuses should distinguish compatible (`0`), incompatible (`1`), and I/O/usage failure (`2`). Normal parsing should remain unchanged; the thorough preflight is opt-in.

## Issue-report workflow and privacy

On incompatibility, text output may offer a pre-filled GitHub issue-draft URL. The default URL must contain diagnostics and environment metadata only—no file path, source, or snippet—and constructing it must not open a browser, send a request, or create an issue.

The initial report should explicitly mention an opt-in rerun flag that can generate a second source-bearing URL when an exact, bounded failed-block reproduction can be extracted. That path must:

* warn that source will appear in terminal logs, browser history, and the GitHub draft;
* include both blocks for duplicate-key failures;
* refuse arbitrary truncation that could make the reproduction invalid;
* use a Markdown fence that source backticks cannot escape;
* refuse overlong issue URLs; and
* reread and hash the file so source from a changed file cannot be attached to a stale report.

The user still reviews the draft and chooses whether to submit it.

## Scope

The preflight is a same-version confidence check, not formal verification, schema validation, or a live provider/LLM feature. It must be deterministic, network-free, read-only, and covered entirely by synthetic tests.

claell:
## Summary

* Add immutable, JSON-serializable compatibility reports for files and strings.
* Check raw-source coverage, explicit parse failures, ordered semantic round trips, canonical stability, output encoding, and informational exact byte identity.
* Add `bibtexparser check FILE` and `python -m bibtexparser check FILE` with text/JSON output and distinct exit statuses.
* Offer a privacy-safe issue draft by default and a separately warned, opt-in source-bearing draft when an exact bounded reproduction is available.
* Reject source-bearing drafts if the file changed after the compatibility report was produced.

Closes #596.

## Scope and commit structure

This additive feature is deliberately split into six commits: core report API, CLI, initial documentation, opt-in reproduction extraction, disclosure documentation, and stale-file protection. Normal parsing and writing behavior is unchanged; the extra parse/write cycles only occur when the preflight is explicitly invoked.

The branch starts directly at upstream `main` and does not require the other open parser, writer, failed-block-policy, Biber-comment, or round-trip-contract PRs. In particular, malformed brace-delimited input is used for failure fixtures so this PR does not depend on #571.

PR #595 currently has a test-local semantic signature. If both proposals proceed, it should be rebased to use the production inventory introduced here so the runtime checker and regression contract have one implementation; that integration is not hidden in this standalone branch.

## Privacy and limits

The default issue URL contains no path or bibliography source and performs no network request. Source inclusion requires `--include-source-in-issue-link`, prints an explicit warning, refuses truncation and overlong URLs, uses a safe Markdown fence, and verifies the reread file hash before constructing the draft.

A passing report means the named checks passed with this parser version. It is not formal verification or an independent oracle for interpretation correctness.

## Validation

* `pytest -p no:cacheprovider -W error tests/test_compatibility.py tests/test_cli.py`: 19 passed.
* Complete upstream suite: 2,595 passed, 12 skipped, with four existing deprecation warnings in `tests/test_entrypoint.py`.
* The full suite under `-W error` reaches only the same four pre-existing warning assertions addressed separately by draft PR #582.

## AI assistance and review note

This pull request was prepared with ChatGPT Codex using GPT-5.6 Sol with high reasoning effort. Codex assisted with contract design, implementation, dependency reduction, privacy review, fixture adaptation, and validation. The work was developed in a concentrated session and has not been field-tested over a long period; automated validation is not a substitute for careful maintainer review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,19 @@ output remains representable in the selected encoding. It reports exact source
 byte identity separately because normal writing may intentionally normalize
 layout. Use `--json` for applications.
 
-An incompatible result includes a reviewable GitHub issue-draft URL by default.
-Constructing the URL does not open a browser, make a network request, or create
-an issue. No file path, bibliography content, or source snippet is included;
-users must review the draft and add only a safe minimal reproduction. Use
-`--no-issue-link` to suppress it.
+An incompatible result includes a reviewable, privacy-safe GitHub issue-draft
+URL by default. Constructing the URL does not open a browser, make a network
+request, or create an issue. No file path, bibliography content, or source
+snippet is included.
+
+The same report explains how to rerun with
+`--include-source-in-issue-link`. That explicit opt-in retains the privacy-safe
+draft and adds a second draft containing an exact failed block when the checker
+can extract a bounded reproduction. For duplicate keys, it includes both
+relevant blocks. It never truncates a block into a potentially invalid example.
+The second URL exposes bibliography source in terminal logs, browser history,
+and the GitHub draft, so review all three before submission. Use
+`--no-issue-link` to suppress issue links entirely.
 
 A passing result means that the file passed the named checks with this version;
 it is not a proof that no unanticipated interpretation bug can exist.
--- a/README.md
+++ b/README.md
@@ -69,6 +69,32 @@ new_bibtex_string = bibtexparser.write_string(bib_database,
 These examples really only show the bare minimum.
 Consult the documentation for a list of available middleware, parsing options and write-formatting options.
 
+## Read-only compatibility preflight
+
+Before trusting a new or unusual bibliography file, exercise the default codec
+contract without modifying the source:
+
+```bash
+python -m bibtexparser check references.bib
+# or, after installation:
+bibtexparser check references.bib
+```
+
+The check verifies raw-source coverage, explicit parse failures, semantic
+parse/write/reparse preservation, canonical-output stability, and whether the
+output remains representable in the selected encoding. It reports exact source
+byte identity separately because normal writing may intentionally normalize
+layout. Use `--json` for applications.
+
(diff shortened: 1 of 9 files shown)


MiWeiss:

PR #599, opened 2026-08-17
:bug: Make InvalidNameError copyable and picklable

eeshsaxena:
`InvalidNameError` can't be copied or pickled, which crashes normal parsing:

```python
import copy
from bibtexparser.middlewares.names import InvalidNameError
copy.deepcopy(InvalidNameError(name="Doe, John", reason="Too many commas"))
# TypeError: InvalidNameError.__init__() missing 1 required positional argument
```

Its `__init__` takes `name` and `reason` but the base `ValueError` only keeps the formatted message in `args`, so deepcopy/pickle rebuild it with a single argument. This surfaces through the public API: a failed name is stored in a `MiddlewareErrorBlock`, and a later middleware like `SortBlocksByTypeAndKeyMiddleware` deepcopies the whole library, so any bad author name plus that middleware raises `TypeError` out of `parse_string`. I store `name`/`reason` and add `__reduce__` so it round-trips. Added a test.

### DIFF
--- a/bibtexparser/middlewares/names.py
+++ b/bibtexparser/middlewares/names.py
@@ -27,9 +27,6 @@ class InvalidNameError(ValueError):
         super().__init__(message)
 
     def __reduce__(self):
-        # Rebuild from name and reason so the exception stays copyable and
-        # picklable; the base ValueError stores only the formatted message in
-        # args, which does not match this two-argument __init__.
         return (self.__class__, (self.name, self.reason))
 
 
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Make InvalidNameError copyable and picklable`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining python-bibtexparser!

MiWeiss:
Thanks a lot for your contribution!

MiWeiss (approved):

PR #601, opened 2026-09-02
🐛 Fix RecursionError on long runs of consecutive newlines

MiWeiss:
`Splitter._next_mark` skips newline marks by recursing, one stack frame per newline.

```python
>>> bibtexparser.parse_string("\n" * 1000 + "@article{a, title={x}}")
RecursionError
```

Hits any file with ~1000+ consecutive lines free of `{ } " , =`. Inside a braced value the error escapes to the caller instead of becoming a failed block.

Fix: loop instead of recurse. Line counting and EOF handling unchanged.

Tests: 6 cases at 3000 newlines, all fail before. Suite 2582 passed, from 2576.

---

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -102,7 +102,6 @@ class Splitter:
             self._current_char_index = m.start()
             return m
 
-        # Iterative: long newline runs would exhaust the recursion limit.
         while True:
             m = next(self._markiter, None)
             if m is None:
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -102,22 +102,24 @@ class Splitter:
             self._current_char_index = m.start()
             return m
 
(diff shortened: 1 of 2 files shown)


bibtexparser/splitter.py
```diff
@@ -102,22 +102,24 @@ def _next_mark(self, accept_eof: bool) -> re.Match | None:
             self._current_char_index = m.start()
             return m
 
-        # Get next mark from iterator
-        m = next(self._markiter, None)
-        if m is not None:
+        # Iterative: long newline runs would exhaust the recursion limit.
```
MiWeiss: Remove this comment

MiWeiss: _claude_

Removed the comment in e918d3b. Suite is green (2583 passed, 12 skipped).

MiWeiss:
lgtm

PR #602, opened 2026-09-02
🐛 Abort entry parsing on a dangling field key before `}`

MiWeiss:
`_move_to_end_of_entry` returns on `}` without checking the text back to the last comma, and the mark regex never matches bare identifiers. A dangling key is dropped silently.

```python
>>> lib = bibtexparser.parse_string("@article{b, title={x}, year}")
>>> lib.entries[0].fields_dict.keys(), lib.failed_blocks
(dict_keys(['title']), [])   # `year` gone, nothing reported
```

Fix: treat `}` as end of entry only when the text back to `key_start` is whitespace (the legal trailing-comma form), else abort into a failed block.

The branch that pushes a closing brace back now also advances `key_start`; without that a normal `@article{a, title={x}}` would falsely abort.

Tests: dangling key in three positions, plus six legal forms pinned. Suite 2609 passed, from 2576.

---

### DIFF
--- a/bibtexparser/splitter.py
+++ b/bibtexparser/splitter.py
@@ -230,6 +230,13 @@ class Splitter:
         while True:
             equals_mark = self._next_mark(accept_eof=False)
             if equals_mark.group(0) == "}":
+                dangling_key = self.bibstr[key_start : equals_mark.start()].strip()
+                if dangling_key:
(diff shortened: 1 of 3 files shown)


MiWeiss:
lgtm

PR #603, opened 2026-09-02
🐛 Apply parse stack only to new blocks in parse_string(library=...)

MiWeiss:
`parse_string(library=existing)` runs the parse stack over the *merged* library, re-transforming blocks an earlier call already transformed. The middlewares are not idempotent.

```python
lib = bibtexparser.parse_string('@string{me = "My Name"}\n@article{a, title = {Hello World}}')
lib = bibtexparser.parse_string('@article{bb, title = {Second}}', library=lib)
```

Before: a spurious "RemoveEnclosingMiddleware must not run before..." warning, and output containing `title = Hello World` and `@string{me = My Name}`. Unenclosed, invalid, does not round-trip.

Fix: split into a fresh library, run the stack there, merge with `fail_on_duplicate_key=False`.

Cross-call string resolution works today, so it is kept: tagged deep copies of the caller's `@string` blocks are seeded into the fresh library and filtered out before the merge. Originals untouched.

Worth a decision: `Splitter.split(library=...)` is now unused inside the package. Left public and untouched.

Tests: 12 cases. The parameter had no coverage at all before. Suite 2588 passed, from 2576.

---

### DIFF
--- a/bibtexparser/entrypoint.py
+++ b/bibtexparser/entrypoint.py
@@ -1,16 +1,23 @@
 import codecs
 import warnings
 from collections.abc import Iterable
+from copy import deepcopy
 from typing import Optional
 from typing import TextIO
 
 from .library import Library
+from .middlewares.enclosing import REMOVED_ENCLOSING_KEY
 from .middlewares.middleware import Middleware
 from .middlewares.parsestack import default_parse_stack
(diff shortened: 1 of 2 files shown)


MiWeiss:
lgtm

PR #605, opened 2026-09-02
🐛 Keep month field enclosing when the value is not transformed

MiWeiss:
`_MonthInterpolator.transform_entry` assigns `month.value` even when the month could not be resolved. The `Field.value` setter resets `enclosing`, so the `no-enclosing` demand on a reference is lost and `AddEnclosingMiddleware` braces it into a literal.

```python
>>> lib = bibtexparser.parse_string("@article{a, month = foo}", append_middleware=[MonthLongStringMiddleware()])
>>> bibtexparser.write_string(lib)
'@article{a,\n\tmonth = {foo}\n}\n'
```

Fix: assign only when the value actually changed, comparing type too, since `"1"` to `1` is a real transformation. `_demanded_enclosing` and the metadata line stay outside the guard, because `MonthAbbreviationMiddleware` demands `no-enclosing` for an already-correct `jan`.

`model.py` untouched; the setter's reset is documented behavior.

Tests: 5 cases across all three month middlewares, 3 fail before. Suite 2581 passed, from 2576.

---

### DIFF
--- a/bibtexparser/middlewares/month.py
+++ b/bibtexparser/middlewares/month.py
@@ -27,7 +27,9 @@ class _MonthInterpolator(BlockMiddleware, abc.ABC):
             return entry
 
         new_val, meta = self.resolve_month_field_val(month)
-        month.value = new_val
+        # Assigning `Field.value` resets the `enclosing` demand references rely on.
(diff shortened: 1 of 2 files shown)


MiWeiss:
lgtm

PR #606, opened 2026-09-02
🐛 Preserve enclosing demand in value-transforming middlewares

MiWeiss:
Same root cause as #605, at two more call sites. Middlewares that rewrite a value's representation lose the `no-enclosing` demand through the `Field.value` setter, so references get braced into literals.

```python
>>> lib = bibtexparser.parse_string("@article{a, month = jan}", append_middleware=[LatexEncodingMiddleware()])
>>> bibtexparser.write_string(lib)
'@article{a,\n\tmonth = {jan}\n}\n'
```

Affects `LatexEncodingMiddleware`, `LatexDecodingMiddleware` and the four name middlewares.

Fix: save and restore the enclosing around the assignment. The `NameParts` branch mutates in place and never hits the setter, so it is left alone. `model.py` untouched.

Known gap, out of scope: latex encoding still escapes the content of unenclosed values, so `intro # outro` becomes `intro \# outro`.

Tests: 31 cases, all fail before. Suite 2607 passed, from 2576.

---

### DIFF
--- a/bibtexparser/middlewares/latex_encoding.py
+++ b/bibtexparser/middlewares/latex_encoding.py
@@ -50,7 +50,10 @@ class _PyStringTransformerMiddleware(BlockMiddleware, abc.ABC):
         errors = []
         for field in entry.fields:
             if isinstance(field.value, str):
+                # The value setter resets `enclosing`; only the representation changes here.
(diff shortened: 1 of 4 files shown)


MiWeiss:
lgtm

PR #607, opened 2026-09-02
🐛 Make Library.remove atomic and tolerant of changed block keys

MiWeiss:
`Library.remove` deletes from `_blocks` before the key dicts. Keys are mutable, so a renamed block raises `KeyError` after it has already been dropped.

```python
>>> e = lib.entries[0]; e.key = "renamed"; lib.remove(e)
KeyError: 'a'
>>> len(lib.blocks), sorted(lib.entries_dict)
(1, ['a', 'c'])   # gone from blocks, still in the key dict
```

The documented contract is `ValueError`. `replace` inherits this, and list removal was non-atomic.

Fix: resolve all indices and key registrations first, mutate after. Registrations are found by identity, falling back to the key, so renames de-register correctly. `remove` is now atomic, mirroring `add`.

Tests: 6 cases, 4 fail before. Suite 2582 passed, from 2576.

---

### DIFF
--- a/bibtexparser/library.py
+++ b/bibtexparser/library.py
@@ -115,6 +115,7 @@ class Library:
             return None
 
         def is_claimed(key: str) -> bool:
+            """Whether this key was already claimed by an earlier block in this call."""
             return any(d is by_key and k == key for d, k in claimed)
 
         for key, registered in by_key.items():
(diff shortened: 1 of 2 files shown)


MiWeiss:
lgtm

PR #608, opened 2026-09-02
🐛 Spread block hashes to avoid pathological set/dict collisions

MiWeiss:
`Block.__hash__` uses `(type, start_line, raw)`, both `None` for blocks built in code, so every such block of a type shares one hash. `set()` then falls back to `__eq__`, which compares the full `__dict__`.

`set()` over N distinct entries:

| N | before | after |
|---|--------|-------|
| 500 | 0.033 s | 0.0001 s |
| 2000 | 0.375 s | 0.0004 s |

Fix: mix in discriminators already compared by `__eq__`, namely entry type and key, string key, preamble and comment text. Values, fields and metadata stay out, as they can be unhashable after middleware. Error-path blocks stay on the base hash; their `__eq__` includes an `Exception`, so instances are never equal anyway.

Tests: 5 cases, including the equal-implies-equal-hash invariant and unhashable values. Suite 2581 passed, from 2576.

---

### DIFF
--- a/bibtexparser/model.py
+++ b/bibtexparser/model.py
@@ -85,6 +85,7 @@ class Block(abc.ABC):
         # Hash on a stable subset of the attributes compared in `__eq__`
         # (equal blocks thus have equal hashes). Mutable or potentially
         # unhashable attributes (e.g. fields, parser_metadata) are excluded.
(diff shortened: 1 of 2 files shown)


MiWeiss:
lgtm

PR #617, opened 2026-09-03
🐛 Materialize iterable middleware arguments before inspecting them

### LINKED ISSUE #531: Iterator passed as append_middleware/prepend_middleware is silently dropped
The parameters are typed `Iterable[Middleware]`, but `_build_parse_stack` consumes the iterable when computing `append_stack_types` before calling `list(append_middleware)` (`bibtexparser/entrypoint.py:36-45`; same pattern in `_build_unparse_stack`, lines 66-75). A generator/iterator therefore results in an empty middleware list — silently.

```python
lib = bibtexparser.parse_string("@article{x, month = {3}}",
    append_middleware=iter([MonthIntMiddleware()]))
lib.entries[0]["month"]  # '3' -- middleware was never applied
```

Fix: materialize to a list at the top of the builders (or type the params as `List`).

MiWeiss:
`append_middleware`/`prepend_middleware` are typed `Iterable[Middleware]`, but the stack builders iterate the argument twice — once to compute middleware types for the duplicate warning, once to build the stack. A generator or other one-shot iterator is exhausted by the first pass, so the middleware is silently dropped.

```python
import bibtexparser
from bibtexparser.middlewares import SeparateCoAuthors

bibtexparser.parse_string('@article{k, author={A and B}}', append_middleware=[SeparateCoAuthors()])
# ['A', 'B']  correct
bibtexparser.parse_string('@article{k, author={A and B}}', append_middleware=iter([SeparateCoAuthors()]))
# 'A and B'   wrong
```

Fix: materialize both arguments to lists at the top of `_build_parse_stack` and `_build_unparse_stack`, before anything inspects them. Public annotations stay `Iterable`; they are now actually honoured.

Tests: 9 regression tests over the public entry points (`parse_string`, `parse_file`, `write_string`, `write_file`) with generators. Suite: 2739 passed, 12 skipped (was 2730 + 12).

Fixes #531

### DIFF
--- a/bibtexparser/entrypoint.py
+++ b/bibtexparser/entrypoint.py
@@ -89,8 +89,8 @@ def _build_unparse_stack(
     stack_types_intersect = parse_stack_types.intersection(append_stack_types)
     if len(stack_types_intersect) > 0:
         warnings.warn(
-            "Some middleware passed in append_middleware are "
-            f"already in the default parse_stack ({stack_types_intersect})."
+            "Some middleware passed in prepend_middleware are "
+            f"already in the default unparse_stack ({stack_types_intersect})."
         )
 
     return list(prepend_middleware) + list(unparse_stack)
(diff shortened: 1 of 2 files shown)


MiWeiss:
lgtm

